PR #76, opened 2024-10-03
[DevOps] feat: llvm cmake headers config fixed

jeremyelalouf:


CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 include(UnixCompileRules)
 include(FetchLLVM)
 
+include_directories(${llvm_BINARY_DIR}/include)
 include_directories(${LLVM_SOURCE_DIR}/include)
```
Snowy1803, line 17: why is that in glu and not under llvm

cmake/FetchLLVM.cmake
```diff
@@ -10,4 +10,8 @@ FetchContent_Declare(
 )
 
 set(LLVM_BUILD_LLVM_DYLIB ON)
+set(LLVM_BUILD_DOCS OFF)
+set(LLDB_INCLUDE_TESTS OFF)
+set(LLVM_ENABLE_PROJECTS "clang;lldb")
```
Snowy1803, line 15: you can also disable most backends

edouard-sn: c'est n'importe quoi jérémy

Snowy1803 (approved):

PR #77, opened 2024-10-04
Create glu::gil::Module class

### LINKED ISSUE #21: Create `glu::gil::Module` class
Top level Module class which contains all declarations in the file.

badermarouan:
Closes #21


include/GIL/Module.hpp
```diff
@@ -0,0 +1,56 @@
...
+    /// @param moduleName A string representing the modules name
+    Module(std::string moduleName);
+
+    /// @brief Adds a Function to the Functions dedicated chained list
+    /// @param function A pointer to the function that will be added
+    void addFunction(Function *function);
+
+    /// @brief Adds a Module to the Modules dedicated chained list
+    /// @param module A pointer to the module that will be added
+    void addModule(Module *module);
```
Snowy1803, line 30: why add a module to a module
explain
(also, imports are planned for October 2025, so next year)

include/GIL/Module.hpp
```diff
@@ -0,0 +1,56 @@
...
+    /// @param moduleName A string that represents the name value
+    /// @return Returns a pointer to the module that corresponds to the name
+    ///         value or NULL if there's no modules that corresponds
+    Module *getModule(std::string moduleName);
+
+    /// @brief Removes the module that corresponds to the given name value
+    /// @param moduleName A string that represents the name value
+    /// @return Returns true if the modules has been removed or false id the
+    ///         Module isn't in the list
+    bool removeModule(std::string moduleName);
```
Snowy1803, line 46: shouldn't be needed, I see no use case for those

include/GIL/Module.hpp
```diff
@@ -0,0 +1,56 @@
...
+/// see https://glu-lang.org/modules/
+class Module {
+public:
+    /// @brief Constructor for a Module instance that initialise its name
+    /// @param moduleName A string representing the modules name
+    Module(std::string moduleName);
+
+    /// @brief Adds a Function to the Functions dedicated chained list
+    /// @param function A pointer to the function that will be added
+    void addFunction(Function *function);
```
Snowy1803, line 26: I think the module should own the functions. Should maybe have unique pointers, OR store the instance directly and have a createFunction instead.

badermarouan: I did it with unique_pointers first but It's not possible with forward declaration, so I guess owning the function is the best way to go.

include/GIL/Module.hpp
badermarouan: I agree but thought that it wouldn't be a bad idea to include it even if we don't see any usages for now.

include/GIL/Module.hpp
badermarouan: okay I'll add an explanation in the oxygen.

include/GIL/Module.hpp
Snowy1803: there should never be a usage for those as it seems like an encapsulation error. the imports are private to the module and shouldn't be queried

include/GIL/Module.hpp
Snowy1803: we will be able to #include <Function> soon.

lib/GIL/Module.cpp
```diff
@@ -0,0 +1,33 @@
...
+{
+    _types.emplace_back(type);
+}
+
+Module *Module::getModule(std::string moduleName)
+{
+    for (auto &it : _modules)
+        if (it->_moduleName == moduleName)
+            return it;
+    return NULL;
```
edouard-sn, line 30: ["Should I use NULL or 0?"](https://www.stroustrup.com/bs_faq2.html#null)

edouard-sn: >  If you have to name the null pointer, call it nullptr; that's what it's called in C++11. Then, "nullptr" will be a keyword.

include/GIL/Module.hpp
```diff
@@ -0,0 +1,49 @@
...
+/// see https://glu-lang.org/modules/
+class Module {
+public:
+    /// @brief Constructor for a Module instance that initialise its name
+    /// @param moduleName A string representing the modules name
+    Module(std::string moduleName);
+
+    /// @brief Adds a Function to the Functions dedicated chained list
+    /// @param function A pointer to the function that will be added
+    void addFunction(Function &&function);
```
Snowy1803, line 24: there's two possibilities:
- you take a unique_ptr so that the function doesn't change its address
- you create the function in place (emplace)
functions must have a stable address, we keep pointers to them all the time. they should have a deleted move ctor

include/GIL/Module.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+    /// @brief Adds a Type to the Types dedicated chained list
+    /// @param type A pointer to the type that will be added
+    void addType(Type *type);
+
+    /// @brief Finds and get the module that corresponds to the given name value
+    /// @param moduleName A string that represents the name value
+    /// @return Returns a pointer to the module that corresponds to the name
+    ///         value or nullptr if there's no modules that corresponds
+    Module *getModule(std::string moduleName);
```
Snowy1803, line 39: still no

include/GIL/Module.hpp
badermarouan: for now I will do the unique pointer method and might add the emplace later after Function class is done

lib/GIL/Module.cpp
```diff
@@ -0,0 +1,15 @@
...
+
+namespace glu::gil {
+
+Module::Module(std::string moduleName) : _moduleName(moduleName)
+{
+}
+
+void Module::addFunction(std::unique_ptr<Function> function)
+{
+    _declarations.emplace_back(std::move(function));
```
Snowy1803, line 12: function should hold ref to this module
function->parent = this;

include/GIL/Module.hpp
Snowy1803: I think emplacing makes more sense, we don't want Functions that aren't inside a module

Snowy1803: the best would be if Function's ctor is private, so they can only be created by modules

badermarouan: I agree but we'll have to wait for the functions PR to be done I close this PR for now

include/GIL/Module.hpp
```diff
@@ -0,0 +1,77 @@
...
+    {
+        return &Module::_functions;
+    }
+
+    /// @brief Adds a Function to the Functions dedicated chained list
+    /// @param name A string representing the function name
+    /// @param type A pointer to the function type
+    void addFunction(
+        std::string name, glu::types::FunctionTy *type,
+        std::list<BasicBlock> basicBlocks
```
Snowy1803, line 43: don't take a basic block, those are created after

include/GIL/Module.hpp
```diff
@@ -0,0 +1,77 @@
...
+    // defined to be used by ilist
+    static FunctionListType Module::*getSublistAccess(Function *)
+    {
+        return &Module::_functions;
+    }
+
+    /// @brief Adds a Function to the Functions dedicated chained list
+    /// @param name A string representing the function name
+    /// @param type A pointer to the function type
+    void addFunction(
```
Snowy1803, line 41: return pointer to the function

lib/GIL/Module.cpp
```diff
@@ -0,0 +1,39 @@
...
+    for (auto &f : _functions)
+        if (f.getName() == name)
+            return &f;
+    return nullptr;
+};
+
+void Module::deleteFunction(Function *f)
+{
+    _functions.remove(f);
+};
```
Snowy1803, line 28: delete f?

lib/GIL/Function.cpp
```diff
@@ -1,4 +1,5 @@
 #include "Function.hpp"
```
LindonAliu, line 1: delete the function.hpp include, is redundant bc its included by module.hpp

lib/GIL/Module.cpp
```diff
@@ -0,0 +1,39 @@
...
+    for (auto &bb : basicBlocks)
+        f->addBasicBlockAtEnd(&bb);
+    _functions.push_back(std::move(f));
+};
+
+Function const *Module::getFunction(std::string name) const
+{
+    for (auto &f : _functions)
+        if (f.getName() == name)
+            return &f;
```
LindonAliu, line 21: for bracket pls

lib/GIL/Module.cpp
```diff
@@ -0,0 +1,39 @@
...
+    _functions.remove(f);
+};
+
+void Module::deleteFunction(std::string name)
+{
+    for (auto &f : _functions)
+        if (f.getName() == name) {
+            _functions.remove(&f);
+            return;
+        }
```
LindonAliu, line 36: for bracket pls

lib/GIL/Module.cpp
badermarouan: when you call remove() it calls delete_node() which already deletes the function f

Snowy1803 (dismissed):

jeremyelalouf (approved):

jeremyelalouf (approved):

PR #79, opened 2024-10-06
feat : add basic file and source management structures

jeremyelalouf:


include/Basic/FileEntry.hpp
```diff
@@ -0,0 +1,98 @@
...
+
+    bool isNamedPipe() const
+    {
+        return _isNamedPipe;
+    }
+
+private:
+    llvm::StringRef _filePath;
+    unsigned int _fileSize = 0;
+    time_t _modificationTime = 0;
```
Snowy1803, line 43: why do we need mtime

include/Basic/SourceLocation.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+private:
+    friend class SourceManager;
+
+    static FileID get(int id)
+    {
+        FileID fid;
+        fid.id = id;
+        return fid;
+    }
```
Snowy1803, line 44: ?? useless? create a private ctor

include/Basic/SourceLocation.hpp
```diff
@@ -0,0 +1,90 @@
...
+        return id > other.id;
+    }
+    bool operator<=(FileID const &other) const
+    {
+        return id <= other.id;
+    }
+    bool operator>=(FileID const &other) const
+    {
+        return id >= other.id;
+    }
```
Snowy1803, line 34: ?? useless

include/Basic/SourceLocation.hpp
```diff
@@ -0,0 +1,90 @@
...
+    int id = 0;
+};
+
+class SourceManager;
+
+class SourceLocation {
+    friend class SourceManager;
+
+private:
+    uint32_t _offset = 0;
```
Snowy1803, line 56: source location = fileid + offset

include/Basic/SourceLocation.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+private:
+    // The offset is an opaque value that is used to refer to a specific
+    // location in the source code. It is not intended to be used directly by
+    // clients of the library, but is instead passed to other functions that
+    // interpret it.
+    uint32_t getOffset() const
+    {
+        return _offset;
+    }
```
Snowy1803, line 86: useless

include/Basic/SourceManager.hpp
```diff
@@ -0,0 +1,195 @@
...
+
+#include <llvm/ADT/SmallVector.h>
+#include <llvm/Support/MemoryBuffer.h>
+#include <optional>
+
+#include "FileEntry.hpp"
+#include "SourceLocation.hpp"
+#include "llvm/ADT/StringRef.h"
+
+namespace glu::srcmanager {
```
Snowy1803, line 12: namespace glu {} for the SourceManager
you can use an anonymous namespace for the rest, if it is only used within the file

include/Basic/SourceManager.hpp
```diff
@@ -0,0 +1,195 @@
...
+
+/// This object owns the MemoryBuffer object containing a file contents.
+class alignas(8) ContentCache {
+
+public:
+    std::optional<FileEntry> _fileEntry;
```
Snowy1803, line 18: why do you have fields at the top and some at the bottom

include/Basic/SourceManager.hpp
```diff
@@ -0,0 +1,195 @@
...
+    {
+        FileInfo info;
+
+        info._includeLoc = includeLoc;
+        info._contentCache = &contentCache;
+        contentCache.setFileName(fileName);
+        return info;
+    }
+
+    SourceLocation getIncludeLoc() const
```
Snowy1803, line 97: what is an include loc??

include/Basic/SourceManager.hpp
```diff
@@ -0,0 +1,195 @@
...
+        SourceLocEntry entry;
+        entry._offset = offset;
+        entry._isExpansion = false;
+        entry._file = file;
+        return entry;
+    }
+
+private:
+    uint32_t _offset;
+    uint32_t _isExpansion : 1;
```
Snowy1803, line 158: ?? we don't have macros
what's the point of this struct??

include/Basic/SourceManager.hpp
```diff
@@ -0,0 +1,195 @@
...
+    /// Set the file ID for the main source file.
+    void setMainFileID(FileID fid)
+    {
+        _mainFile = fid;
+    }
+
+    FileID getMainFileID() const
+    {
+        return _mainFile;
+    }
```
Snowy1803, line 190: having a "main" file seems dangerous to me but ok

Snowy1803 (commented):
add documentation to all classes, I don't understand why most of them even exist

include/Basic/SourceManager.hpp
```diff
@@ -0,0 +1,273 @@
...
+///
+class alignas(8) ContentCache {
+
+private:
+    llvm::StringRef _fileName;
+
+    /// The buffer containing the content of the file.
+    mutable std::unique_ptr<llvm::MemoryBuffer> _buffer;
+    /// A flag to indicate if the buffer is invalid.
+    mutable unsigned _isBufferInvalid : 1;
```
Snowy1803, line 30: why use a `ContentCache const *` then use `mutable` fields... just stop using const?

Snowy1803: and why not directly use an std::optional... or just set the buffer to nullptr when it's invalid

include/Basic/SourceManager.hpp
```diff
@@ -0,0 +1,273 @@
...
+    /// If the FileInfo is for the main file this location is invalid (ID == 0).
+    SourceLocation _importLoc;
+
+    /// The content cache of the file.
+    ContentCache const *_contentCache;
+
+public:
+    FileInfo()
+        : _importLoc(0)
+        , _contentCache(nullptr) {};
```
Snowy1803, line 107: semicolon shouldn't be there

include/Basic/SourceManager.hpp
```diff
@@ -0,0 +1,273 @@
...
+    get(SourceLocation includeLoc, ContentCache &contentCache,
+        llvm::StringRef fileName)
+    {
+        FileInfo info;
+
+        info._importLoc = includeLoc;
+        info._contentCache = &contentCache;
+        contentCache.setFileName(fileName);
+        return info;
+    }
```
Snowy1803, line 120: why is this not a constructor

include/Basic/SourceManager.hpp
```diff
@@ -0,0 +1,273 @@
...
+    SourceLocEntry(uint32_t offset, FileInfo const &file)
+        : _offset(offset)
+        , _file(file)
+    {
+    }
+
+    /// Returnn a const reference to the FileInfo object.
+    FileInfo const &getFileInfo() const
+    {
+        return const_cast<SourceLocEntry *>(this)->getFile();
```
Snowy1803, line 170:
```suggestion
        return _file;
```
is that not easier??

lib/Basic/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+
+target_link_libraries(gluBasic
+    PUBLIC
+    LLVM
+)
+
+target_sources(gluBasic
+    PRIVATE
+    SourceManager.cpp
+)
```
Snowy1803, line 16: you should remove libTokens, too (you can do that in the tokens PR)

include/Basic/SourceManager.hpp
Snowy1803: I don't even understand why FileInfo and ContentCache are separate

include/Basic/SourceManager.hpp
```diff
@@ -0,0 +1,273 @@
...
+/// interpreting SourceLocEntry objects.
+///
+/// A SourceLocEntry is basicly an offset into the complete source code. The
+/// SourceManager knowings to which file this offset belongs can interpret it
+/// and provide useful informations from it.
+///
+class SourceLocEntry {
+
+private:
+    /// The offset of the source location.
```
Snowy1803, line 154: unclear. this should explain what it's useful for, not just repeat the name of the variable and the name of the class

include/Basic/SourceManager.hpp
```diff
@@ -0,0 +1,273 @@
...
+/// @class SourceLocEntry
+/// @brief Represents a specific location in the source code.
+///
+/// A SourceLocEntry is an object that represents a specific location
+/// in the source code. It is used to refer to a specific character in a
+/// specific file. The SourceManager class is responsible for creating and
+/// interpreting SourceLocEntry objects.
+///
+/// A SourceLocEntry is basicly an offset into the complete source code. The
+/// SourceManager knowings to which file this offset belongs can interpret it
```
Snowy1803, line 148:
```suggestion
/// SourceManager knows which file an offset belong to, and can interpret it
```
lots of typos too

include/Basic/SourceManager.hpp
```diff
@@ -0,0 +1,273 @@
...
+/// The SourceManager knows how to interpret a SourceLocation object and provide
+/// information about the source code at that location in an efficient way.
+///
+class SourceManager {
+
+private:
+    /// Every source location of the source code.
+    llvm::SmallVector<SourceLocEntry, 0> _sourceLocs;
+    /// The content caches of the loaded files.
+    std::vector<ContentCache *> _contentCaches;
```
Snowy1803, line 204: why
they're already in the above vector, why do you need a second vector?
+ it's using std vector

include/Basic/SourceManager.hpp
```diff
@@ -0,0 +1,236 @@
...
+///
+/// @class ContentCache
+/// @brief A file loaded will be cached in this class.
+///
+/// The ContentCache class is responsible for caching the content of a file that
+/// has been loaded. It is used by the SourceManager to store the content of a
+/// file. The content of a file is stored in a MemoryBuffer object which is
+/// owned by the ContentCache.
+///
+class alignas(8) ContentCache {
```
Snowy1803, line 22: why alignas?

include/Basic/SourceManager.hpp
```diff
@@ -0,0 +1,236 @@
...
+
+public:
+    /// Create a FileInfo object with the given SourceLocation and ContentCache.
+    FileInfo(
+        SourceLocation includeLoc, ContentCache &contentCache,
+        llvm::StringRef fileName
+    )
+        : _importLoc(includeLoc), _contentCache(nullptr)
+    {
+        _contentCache = &contentCache;
```
Snowy1803, line 103: why do you take a reference??
new -> dereference -> take address -> put in unique ptr
why not just use make_unique and take the unique_ptr as parameter
or better, put it directly in the struct as you don't need an indirection? (your types can all be move-only)

include/Basic/SourceManager.hpp
```diff
@@ -0,0 +1,236 @@
...
+        contentCache.setFileName(fileName);
```
Snowy1803, line 104: why is the filename in the content cache and not in the FileInfo?

include/Basic/SourceManager.hpp
```diff
@@ -0,0 +1,236 @@
...
+///
+/// The ContentCache class is responsible for caching the content of a file that
+/// has been loaded. It is used by the SourceManager to store the content of a
+/// file. The content of a file is stored in a MemoryBuffer object which is
+/// owned by the ContentCache.
+///
+class alignas(8) ContentCache {
+
+private:
+    llvm::StringRef _fileName;
```
Snowy1803, line 25: this will be a reference to a temporary, which will be deallocated. it will crash. use std::string to keep a copy

lib/Basic/SourceManager.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+    SourceLocEntry entry(
+        newOffset, FileInfo::get(SourceLocation(0), *cache, *fileName)
+    );
+
+    if (_sourceLocs.empty()) {
+        _mainFile = FileID(_sourceLocs.size());
+    }
+
+    _sourceLocs.push_back(entry);
```
Snowy1803, line 41: std::move?
to me all of those classes should be move-only, and I still don't understand why SourceLocEntry, FileInfo, and ContentCache are not just one class

include/Basic/SourceLocation.hpp
```diff
@@ -0,0 +1,73 @@
...
+/// SourceManager.
+///
+class FileID {
+    friend class SourceManager;
+
+private:
+    /// The opaque identifier for the file.
+    int _id = 0;
+
+private:
```
Snowy1803, line 22: why write private if everything is private? it's a class so everything is private by default

include/Basic/SourceLocation.hpp
```diff
@@ -0,0 +1,73 @@
...
+    /// the location cannot point to a valid file, line, or column in the code.
+    bool isValid() const { return _offset != 0; }
+    bool isInvalid() const { return _offset == 0; }
+
+private:
+    /// The offset is an opaque value that is used to refer to a specific
+    /// location in the source code. It is not intended to be used directly by
+    /// clients of the library, but is instead passed to other functions (mostly
+    /// SourceManager's functions) that interpret it.
+    uint32_t getOffset() const { return _offset; }
```
Snowy1803, line 69: why make a private getter? in private context, you can directly use `_offset`. getters are for when you need the `get` operation to be public (vs a private `set`)

Snowy1803 (approved):

PR #96, opened 2024-10-09
Add TerminatorInst

### LINKED ISSUE #26: Create `glu::gil::TerminatorInst` base class
instruction at the end of a basic block. inherits from GILInstruction, is implemented by 4 other classes

Snowy1803:
Closes #26

### DIFF
--- a/lib/GIL/BasicBlock.cpp
+++ b/lib/GIL/BasicBlock.cpp
@@ -6,12 +6,11 @@ namespace glu::gil {
 
 InstBase *BasicBlock::popFirstInstruction()
 {
-    if (!_instructions.empty()) {
-        InstBase *inst = &_instructions.front();
-        _instructions.remove(inst);
-        return inst;
-    }
-    return nullptr;
+    if (_instructions.empty())
+        return nullptr;
+    InstBase *inst = &_instructions.front();
+    _instructions.remove(inst);
+    return inst;
 }
 
 void BasicBlock::addInstructionBefore(InstBase *inst, InstBase *before)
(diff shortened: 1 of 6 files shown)


jeremyelalouf (dismissed):

LindonAliu (dismissed):
Let me merge my pr before yours, so that afterwards you can uncomment the commented isTerminator in BasicBlock and delete the placeholders in InstBase

lib/GIL/BasicBlock.cpp
```diff
@@ -53,28 +52,16 @@ void BasicBlock::removeInstruction(InstBase *inst)
...
-    if (terminator) {
-        return static_cast<TerminatorInst *>(terminator);
-    }
-    return nullptr;
+    if (_instructions.empty())
+        return nullptr;
+    return llvm::dyn_cast<TerminatorInst>(&_instructions.back());
 }
 
 void BasicBlock::setTerminator(InstBase *terminator)
```
LindonAliu, line 62: maybe is better to have an argument that is a `TerminatorInst *` and not an `InstBase *` ?

lib/GIL/BasicBlock.cpp
```diff
@@ -53,28 +52,16 @@ void BasicBlock::removeInstruction(InstBase *inst)
...
 {
-    InstBase *lastInst = getTerminator();
-    bool lastInstIsTerminator = lastInst != nullptr;
-
-    if (lastInstIsTerminator) {
+    if (TerminatorInst *lastInst = getTerminatorInst()) {
```
LindonAliu, line 64: I don't think it's pretty

lib/GIL/BasicBlock.cpp
```diff
@@ -53,28 +52,16 @@ void BasicBlock::removeInstruction(InstBase *inst)
...
 {
-    InstBase *terminator = getTerminator();
-    if (terminator) {
-        return static_cast<TerminatorInst *>(terminator);
-    }
-    return nullptr;
+    if (_instructions.empty())
+        return nullptr;
+    return llvm::dyn_cast<TerminatorInst>(&_instructions.back());
 }
```
LindonAliu, line 60: no use of the `isTerminator()` function ?

lib/GIL/BasicBlock.cpp
Snowy1803: why write 4 lines when it can be 1 line? that's just the syntax to unwrap an optional / check for null in C++

lib/GIL/BasicBlock.cpp
Snowy1803: isTerminator does a llvm::isa (which returns a boolean), llvm::dyn_cast returns a pointer or null. why do the check twice?

LindonAliu (approved):

jeremyelalouf (approved):

PR #103, opened 2024-10-15
feat: ASTNode base class with NodeKind

jeremyelalouf:


include/AST/ASTNode.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+    /// The location from which the node was parsed.
+    SourceLocation _nodeLocation;
+
+    /// The AST node kind of the current node.
+    NodeKind _nodeKind;
+
+public:
+    ASTNode();
+    ~ASTNode() = default;
```
Snowy1803, line 39: take kind as argument, add getter (see https://llvm.org/docs/HowToSetUpLLVMStyleRTTI.html)

include/AST/ASTNode.hpp
```diff
@@ -0,0 +1,42 @@
+#include "SourceLocation.hpp"
```
Snowy1803, line 1: include guard

include/AST/NodeKind.def
```diff
@@ -0,0 +1,49 @@
+#ifndef NODE_KIND
+    #error "NODE_KIND must be defined"
+    #define NODE_KIND(Name)
```
Snowy1803, line 3: will need parent here too, and everywhere (can't implement the visitor otherwise)

include/AST/ASTNode.hpp
```diff
@@ -0,0 +1,42 @@
...
+ * @brief The kind of a node in the AST.
+ **/
+enum class NodeKind {
+#define NODE_KIND(Name) Name##Kind,
+#define NODE_KIND_SUPER(Name, Parent) Name##FirstKind,
+#define NODE_KIND_SUPER_END(Name) Name##LastKind,
+#include "NodeKind.def"
+#undef NODE_KIND
+#undef NODE_KIND_SUPER
+#undef NODE_KIND_SUPER_END
```
Snowy1803, line 15: you already undef at the end of NodeKind.def, so no need

Snowy1803 (approved):

yunsecode (approved):

PR #105, opened 2024-10-15
Feature: add `FunctionTy` class in AST

### LINKED ISSUE #90: Create `glu::types::FunctionTy` (Function type)
Values can't directly have this type, but can have PointerTy to FunctionTy

LindonAliu:
closes #90

### DIFF
--- a/include/AST/Types/FunctionTy.hpp
+++ b/include/AST/Types/FunctionTy.hpp
@@ -9,7 +9,7 @@ namespace glu::types {
 /// @brief FunctionTy is a class that represents a function type in the AST.
 class FunctionTy : public TypeBase {
     std::vector<TypeBase *> const _parameters;
-    TypeBase *_returnType;
+    TypeBase *const _returnType;
 
 public:
     /// @brief Constructor for the FunctionTy class.
--- a/include/AST/Types/FunctionTy.hpp
+++ b/include/AST/Types/FunctionTy.hpp
@@ -8,7 +8,6 @@ namespace glu::types {
 
 /// @brief FunctionTy is a class that represents a function type in the AST.
 class FunctionTy : public TypeBase {
-private:
     std::vector<TypeBase *> const _parameters;
     TypeBase *_returnType;
 
--- a/include/AST/Types/FunctionTy.hpp
+++ b/include/AST/Types/FunctionTy.hpp
@@ -18,7 +18,7 @@ public:
     /// parameters of the function.
     /// @param returnType A TypeBase pointer representing the return type of the
     /// function.
-    FunctionTy(std::vector<TypeBase *> const &&parameters, TypeBase *returnType)
+    FunctionTy(std::vector<TypeBase *> parameters, TypeBase *returnType)
         : TypeBase(TypeKind::FunctionTyKind)
         , _parameters(std::move(parameters))
         , _returnType(returnType)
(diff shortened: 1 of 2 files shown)


include/AST/Types/FunctionTy.hpp
```diff
@@ -0,0 +1,48 @@
...
+    std::vector<TypeBase *> const _parameters;
+    TypeBase *_returnType;
+
+public:
+    /// @brief Constructor for the FunctionTy class.
+    /// @param parameters A vector of TypeBase pointers representing the
+    /// parameters of the function.
+    /// @param returnType A TypeBase pointer representing the return type of the
+    /// function.
+    FunctionTy(std::vector<TypeBase *> const &&parameters, TypeBase *returnType)
```
Snowy1803, line 21: what is the point of a const rvalue reference??
better take an owned value, but rvalue-ref works too

include/AST/Types/FunctionTy.hpp
```diff
@@ -0,0 +1,48 @@
...
+    /// @return Returns true if the TypeBase is a FunctionTy, false otherwise.
+    static bool classof(TypeBase const *type)
+    {
+        return type->getKind() == TypeKind::FunctionTyKind;
+    }
+
+    /// @brief Getter for the parameters of the function.
+    /// @return Returns a vector of TypeBase pointers representing the
+    /// parameters
+    std::vector<TypeBase *> const &getParameters() const { return _parameters; }
```
Snowy1803, line 39: hide implementation details
have getParameter(int index) and getParameterCount()

include/AST/Types/FunctionTy.hpp
```diff
@@ -0,0 +1,56 @@
...
+public:
+    /// @brief Constructor for the FunctionTy class.
+    /// @param parameters A vector of TypeBase pointers representing the
+    /// parameters of the function.
+    /// @param returnType A TypeBase pointer representing the return type of the
+    /// function.
+    FunctionTy(std::vector<TypeBase *> parameters, TypeBase *returnType)
+        : TypeBase(TypeKind::FunctionTyKind)
+        , _parameters(std::move(parameters))
+        , _returnType(returnType)
```
badermarouan, line 24: I think this should be a kind because we don't know about the return value when the class is being instantiate, only the return type

include/AST/Types/FunctionTy.hpp
```diff
@@ -0,0 +1,56 @@
...
+    /// @return Returns true if the TypeBase is a FunctionTy, false otherwise.
+    static bool classof(TypeBase const *type)
+    {
+        return type->getKind() == TypeKind::FunctionTyKind;
+    }
+
+    /// @brief Getter for a specific parameter of the function.
+    /// @param index The index of the parameter to get.
+    /// @return Returns a TypeBase pointer representing the parameter
+    TypeBase *getParameter(std::size_t index) const
```
badermarouan, line 39: An overload of [] operator could be useful

include/AST/Types/FunctionTy.hpp
```diff
@@ -0,0 +1,56 @@
...
+#define GLU_AST_TYPES_FUNCTIONTY_HPP
+
+#include "TypeBase.hpp"
+#include <vector>
+
+namespace glu::types {
+
+/// @brief FunctionTy is a class that represents a function type in the AST.
+class FunctionTy : public TypeBase {
+private:
```
badermarouan, line 11: Not needed it is implicit in class, we didn't explicitly declared it in other classes

include/AST/Types/FunctionTy.hpp
LindonAliu: No, it's not parameter encapsulation

include/AST/Types/FunctionTy.hpp
LindonAliu: no defined rules for that at the moment, we can discuss it :)

badermarouan (dismissed):

yunsecode (dismissed):

include/AST/Types/FunctionTy.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+#include "TypeBase.hpp"
+#include <vector>
+
+namespace glu::types {
+
+/// @brief FunctionTy is a class that represents a function type in the AST.
+class FunctionTy : public TypeBase {
+    std::vector<TypeBase *> const _parameters;
+    TypeBase *_returnType;
```
Snowy1803, line 12: if params are const why not return type?
```suggestion
    TypeBase *const _returnType;
```

yunsecode (approved):

PR #144, opened 2024-11-26
Feature: add LetDecl and VarDecl in AST

### LINKED ISSUE #59: Implement `gil::ast::VarDecl` class

### LINKED ISSUE #60: Implement `gil::ast::LetDecl` class

LindonAliu:
fix #59 #60


LindonAliu:
Maybe this class is the same as #59 class, what do you think to do something more generic ?

include/AST/LetDecl.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+/// @class LetDecl
+/// @brief Represents a let declaration in the AST.
+///
+/// This class inherits from DeclBase and encapsulates the details of a 'let'
+/// declaration, including its name, type, and value.
+class LetDecl : public DeclBase {
+    std::string _name;
+    glu::types::TypeBase *_type;
+    glu::ast::ASTNode *_value;
```
Snowy1803, line 19: `ExprBase *`?

include/AST/LetDecl.hpp
```diff
@@ -0,0 +1,54 @@
...
+    /// @brief Constructor for the LetDecl class.
+    /// @param location The source location of the declaration.
+    /// @param parent The parent AST node.
+    /// @param name The name of the declared variable.
+    /// @param type The type of the declared variable.
+    /// @param value The value assigned to the declared variable.
+    LetDecl(
+        SourceLocation location, ASTNode *parent, std::string name,
+        glu::types::TypeBase *type, ASTNode *value
+    )
```
Snowy1803, line 31: you can't put both parent and children in the constructor... we either create parents then children, or children then parents. Creating the children and setting the parent in the constructor is probably easier, so never set the parent in constructors?

include/AST/LetDecl.hpp
```diff
@@ -0,0 +1,54 @@
...
+        : DeclBase(NodeKind::LetDeclKind, location, parent)
+        , _name(std::move(name))
+        , _type(type)
+        , _value(value)
+    {
```
Snowy1803, line 36: like
```suggestion
    {
        value->setParent(this);
```

include/AST/LetDecl.hpp
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    /// @brief Getter for the name of the declared variable.
+    /// @return Returns the name of the declared variable.
+    std::string getName() const { return _name; }
```
Snowy1803, line 41:
```suggestion
    std::string const &getName() const { return _name; }
```

include/AST/LetDecl.hpp
LindonAliu: I agree that the parent of `value` is `this`, but here I need the parent because the class inherits from `DeclBase`, which itself needs a parent, so I don't really have a choice but to use one in the constructor

Snowy1803 (approved):

test/AST/LetDecl.cpp
```diff
@@ -0,0 +1,34 @@
...
+    std::string name = "x";
+    TypeBase *type = new BoolTy();
+    ExprBase *value = new TestExprBase();
+
+    LetDecl letDecl(loc, name, type, value);
+
+    EXPECT_EQ(letDecl.getName(), name);
+    EXPECT_EQ(letDecl.getType(), type);
+    EXPECT_EQ(letDecl.getValue(), value);
+    EXPECT_TRUE(llvm::isa<DeclBase>(letDecl));
```
Snowy1803, line 30: those assertions are quite useless... one more important one is `value.getParent() == &letDecl`

Snowy1803 (approved):

include/AST/Decl/LetDecl.hpp
```diff
@@ -0,0 +1,33 @@
...
+    /// @param value The value assigned to the declared variable.
+    LetDecl(
+        SourceLocation location, std::string name, glu::types::TypeBase *type,
+        ExprBase *value
+    )
+        : VarLetDecl(
+              NodeKind::LetDeclKind, location, std::move(name), type, value
+          )
+    {
+    }
```
Snowy1803, line 28: you still need a classof function here and in VarDecl

jeremyelalouf (approved):

yunsecode (approved):

PR #148, opened 2024-12-02
AllocaInst class

### LINKED ISSUE #35: Create `glu::gil::AllocaInst` class

badermarouan:
Fix #35

### DIFF
--- a/include/GIL/Instructions.hpp
+++ b/include/GIL/Instructions.hpp
@@ -2,6 +2,7 @@
 #ifndef GLU_GIL_INSTRUCTIONS_HPP
 #define GLU_GIL_INSTRUCTIONS_HPP
 
+#include "Instructions/AllocaInst.hpp"
 #include "Instructions/BrInst.hpp"
 #include "Instructions/CallInst.hpp"
 #include "Instructions/CastIntToPtrInst.hpp"
@@ -18,7 +19,6 @@
 #include "Instructions/StringLiteralInst.hpp"
 #include "Instructions/StructExtractInst.hpp"
 #include "Instructions/UnreachableInst.hpp"
-#include "Instructions/AllocaInst.hpp"
 
 // Needed by InstBase
 #include "BasicBlock.hpp"
--- a/include/GIL/Instructions.hpp
+++ b/include/GIL/Instructions.hpp
@@ -18,6 +18,7 @@
 #include "Instructions/StringLiteralInst.hpp"
 #include "Instructions/StructExtractInst.hpp"
 #include "Instructions/UnreachableInst.hpp"
+#include "Instructions/AllocaInst.hpp"
 
 // Needed by InstBase
 #include "BasicBlock.hpp"
--- a/include/GIL/Instructions/AllocaInst.hpp
+++ b/include/GIL/Instructions/AllocaInst.hpp
@@ -21,8 +21,8 @@ public:
     /// @param size The size of the allocation.
     /// @param type The type of the allocation.
     ///
-    AllocaInst(size_t size, Type type)
-        : InstBase(InstKind::AllocaInstKind)
+    AllocaInst(Type type)
+        : InstBase(InstKind::AllocaInstKind), type(type)
     {
     }
 
--- a/include/GIL/Instructions/AllocaInst.hpp
+++ b/include/GIL/Instructions/AllocaInst.hpp
@@ -5,40 +5,40 @@
 
 namespace glu::gil {
 
-/**
- * @class AllocaInst
- * @brief Represents an allocation instruction.
- *
- * This class is derived from InstBase and represents an allocation instruction
- * in the GLU GIL (Generic Intermediate Language).
- */
+///
+///  @class AllocaInst
+///  @brief Represents an allocation instruction.
+///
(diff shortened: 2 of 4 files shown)


include/GIL/Instructions/AllocaInst.hpp
```diff
@@ -0,0 +1,47 @@
...
+namespace glu::gil {
+
+///
+///  @class AllocaInst
+///  @brief Represents an allocation instruction.
+///
+///  This class is derived from InstBase and represents an allocation instruction
+///  in the GLU GIL (Generic Intermediate Language).
+///
+class AllocaInst : public InstBase {
```
Snowy1803, line 15: you are not implementing any of the abstract methods

include/GIL/Instructions/AllocaInst.hpp
```diff
@@ -0,0 +1,47 @@
...
+    size_t size; ///< The size of the allocation.
```
Snowy1803, line 16: there is no size, only a type (the Type type already stores the size anyway)

include/GIL/Instructions/AllocaInst.hpp
```diff
@@ -0,0 +1,61 @@
...
+class AllocaInst : public InstBase {
+    Type type; ///< The type of the allocation.
+public:
+    ///
+    /// @brief Constructs an AllocaInst object.
+    ///
+    /// @param size The size of the allocation.
+    /// @param type The type of the allocation.
+    ///
+    AllocaInst(size_t size, Type type)
```
Snowy1803, line 24: no size

include/GIL/Type.hpp
```diff
@@ -35,6 +36,11 @@ class Type {
         );
     }
 
+    Type(Type const *type) : _fields { type->_fields }
+    {
+        _type = new glu::types::PointerTy(type->_type);
+    }
```
Snowy1803, line 42: 1- you cannot call `new` to create a type, we'll use an arena, so that would break pointer equality rules, and it would never be freed
2 -  the size/alignment of a pointer is not the same as the pointee
3 - why would a fake copy constructor create a pointer type instead of a copy??
4 - this is never called so it doesn't even resolve the problem

Snowy1803 (changes requested):

include/GIL/Type.hpp
badermarouan: This is still work in progress/in discussion that's why I didn't ask for a new review

include/GIL/Instructions.hpp
```diff
@@ -18,6 +18,7 @@
 #include "Instructions/StringLiteralInst.hpp"
 #include "Instructions/StructExtractInst.hpp"
 #include "Instructions/UnreachableInst.hpp"
+#include "Instructions/AllocaInst.hpp"
```
Snowy1803, line 21: clang format

include/GIL/Instructions/AllocaInst.hpp
```diff
@@ -0,0 +1,72 @@
...
+    TypeMemory *_typeMemory; ///< The Typed Memory Arena.
+    Type _ptr; ///< The pointer type.
+public:
+    ///
+    /// @brief Constructs an AllocaInst object.
+    ///
+    /// @param type The type of the allocation.
+    /// @param astTypeMemory The Typed Memory Arena.
+    ///
+    AllocaInst(Type type, TypeMemory *astTypeMemory)
```
Snowy1803, line 30: Take the context

include/GIL/Instructions/AllocaInst.hpp
```diff
@@ -0,0 +1,72 @@
...
+        : InstBase(InstKind::AllocaInstKind)
+        , _type(type)
+        , _typeMemory(astTypeMemory)
```
Snowy1803, line 33: Don’t store it, it’s not used

Snowy1803: still, don't store the context, you don't use it.

include/GIL/Instructions/AllocaInst.hpp
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    virtual size_t getResultCount() const override { return 1; }
+
+    virtual size_t getOperandCount() const override { return 1; }
+
+    virtual Operand getOperand(size_t index) const override
+    {
+        assert(index == 0 && "Operand index out of range");
+        return Operand(_ptr);
```
Snowy1803, line 48: the operand is the pointee type, not the pointer type (we allocate a value, not a pointer)

include/GIL/Instructions/AllocaInst.hpp
```diff
@@ -0,0 +1,65 @@
...
+///
+///  @class AllocaInst
+///  @brief Represents an allocation instruction.
+///
+///  This class is derived from InstBase and represents an allocation
+///  instruction in the GLU GIL (Generic Intermediate Language).
+///
+class AllocaInst : public InstBase {
+    using Context = glu::ast::ASTContext;
+    Context *_context; ///< The AST Context.
```
Snowy1803, line 20: don't store the context, it has nothing to do here

Snowy1803 (approved):

PR #158, opened 2025-01-31
[Lexer] Feat: Add FLEX scanner

### LINKED ISSUE #153: Create Flex code
Create the actual code for the lexical analysis, generating a token stream from a string, using flex

Snowy1803:
Closes #153

### DIFF
--- a/lib/Lexer/CMakeLists.txt
+++ b/lib/Lexer/CMakeLists.txt
@@ -13,7 +13,6 @@ target_link_libraries(Lexer
     PUBLIC
     gluBasic
     LLVM
-    #   ${FLEX_LIBRARIES}
 )
 
 target_sources(Lexer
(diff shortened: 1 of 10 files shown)


lib/Lexer/CMakeLists.txt
```diff
@@ -3,14 +3,20 @@ add_library (Lexer SHARED)
...
+    ${CMAKE_CURRENT_BINARY_DIR}
 )
 
+FLEX_TARGET(Scanner Scanner.l  ${CMAKE_CURRENT_BINARY_DIR}/Scanner.cpp)
+
 target_link_libraries(Lexer
     PUBLIC
     gluBasic
+    LLVM
+    #   ${FLEX_LIBRARIES}
```
jeremyelalouf, line 16: You don't need this ?

Snowy1803: This links with the fl lib. The problem is that lib is just called l on macOS, so it doesn't link.
That library just contains 2 function: a default main (but we have our own), and a default wrap function (but we have wrap disabled in the options within Scanner.l)
So it's useless and it doesn't even work.

jeremyelalouf: delete it

jeremyelalouf (approved):

LindonAliu (approved):

PR #161, opened 2025-01-31
Create FunctionPtrInst class

### LINKED ISSUE #47: Create `glu::gil::FunctionPtrInst` class

badermarouan:
Fix #47

### DIFF
--- a/include/GIL/Instructions/FunctionPtrInst.hpp
+++ b/include/GIL/Instructions/FunctionPtrInst.hpp
@@ -1,0 +1,75 @@
+#ifndef GLU_GIL_FUNCTION_PTR_INST_HPP
+#define GLU_GIL_FUNCTION_PTR_INST_HPP
+
+#include "Function.hpp"
+
+namespace glu::gil {
+
+/// @class FunctionPtrInst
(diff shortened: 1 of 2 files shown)


include/GIL/Instructions/FunctionPtrInst.hpp
```diff
@@ -0,0 +1,75 @@
+#ifndef GLU_GIL_FUNCTION_PTR_INST_HPP
+#define GLU_GIL_FUNCTION_PTR_INST_HPP
+
+#include "Function.hpp"
+
+namespace glu::gil {
+
+/// @class FunctionPtrInst
+/// @brief Represents an instruction to hold a pointer to a function.
```
jeremyelalouf, line 9: To create a pointer to a function is easier to read no ?

jeremyelalouf (approved):

include/GIL/Instructions/FunctionPtrInst.hpp
badermarouan: I think in the context of an instruction in gil "hold a pointer" makes more sense. But we can reformulate into: "Represents an instruction that creates and holds a pointer to a function" if you think it's not precise enough.

PR #164, opened 2025-02-01
Feature/ missing tokens for operators and `@` punctuator

jeremyelalouf:


include/Basic/TokenKind.def
```diff
@@ -109,6 +114,30 @@ PUNCTUATOR(exclamation, '!')
...
+OPERATOR(gt, '>')
+OPERATOR(ge, '>=')
+OPERATOR(and, '&&')
+OPERATOR(or, '||')
+OPERATOR(bitAnd, '&')
+OPERATOR(bitOr, '|')
+OPERATOR(bitXor, '^')
+OPERATOR(bitLShift, '<<')
+OPERATOR(bitRShift, '>>')
+OPERATOR(range, '...')
```
Snowy1803, line 138: remove it from punctuators, it can't be both

include/Basic/TokenKind.def
```diff
@@ -109,6 +114,30 @@ PUNCTUATOR(exclamation, '!')
 PUNCTUATOR(question, '?')
 PUNCTUATOR(stringQuote, '\"')
```
Snowy1803, line 115: this is not a token that exists, they would be string literals

include/Basic/TokenKind.def
```diff
@@ -109,6 +114,30 @@ PUNCTUATOR(exclamation, '!')
...
 PUNCTUATOR(singleQuote, '\'')
```
Snowy1803, line 116: is this in the grammar anywhere?

include/Basic/TokenKind.def
```diff
@@ -109,6 +114,30 @@ PUNCTUATOR(exclamation, '!')
...
+OPERATOR(ge, '>=')
+OPERATOR(and, '&&')
+OPERATOR(or, '||')
+OPERATOR(bitAnd, '&')
+OPERATOR(bitOr, '|')
+OPERATOR(bitXor, '^')
+OPERATOR(bitLShift, '<<')
+OPERATOR(bitRShift, '>>')
+OPERATOR(range, '...')
+OPERATOR(exclusiveRange, '..<')
```
Snowy1803, line 139: you didn't put the unary operators, where's `~`? exclamation should be here and not in punctuators as well

include/Basic/TokenKind.def
```diff
@@ -102,13 +107,34 @@ PUNCTUATOR(comma, ',')
...
+OPERATOR(exclamation, '!')
+OPERATOR(tilde, '~')
```
Snowy1803, line 137: why do all the operators have names based on their usage, but these two have names based on their symbol

not, compl

Snowy1803 (approved):

PR #197, opened 2025-02-17
Memory: Implement type interning

### LINKED ISSUE #196: Implement type interning in MemoryArena
In memoryarena, calling create twice with the same arguments returns the same ptr twice.

LindonAliu:
fix #196

### DIFF
--- a/include/AST/Decl/EnumDecl.hpp
+++ b/include/AST/Decl/EnumDecl.hpp
@@ -16,6 +16,7 @@ namespace glu::ast {
 /// declaration.
 class EnumDecl : public DeclBase {
     using EnumTy = glu::types::EnumTy;
+    using Case = glu::types::Case;
     EnumTy *_self;
 
 public:
@@ -28,7 +29,7 @@ public:
     /// the enum.
     EnumDecl(
         ASTContext &context, SourceLocation location, ASTNode *parent,
-        std::string name, llvm::SmallVector<EnumTy::Case> cases
+        std::string name, llvm::SmallVector<Case> cases
     )
         : DeclBase(NodeKind::EnumDeclKind, location, parent)
         , _self(context.getTypesMemoryArena().create<EnumTy>(
@@ -39,7 +40,7 @@ public:
 
     /// @brief Getter for the name of the enum.
     /// @return Returns the name of the enum.
-    std::string getName() const { return _self->getName(); }
+    llvm::StringRef getName() const { return _self->getName(); }
 
     /// @brief Getter for the type of this enum.
     /// @return Returns the type of this enum.
--- a/include/AST/Decl/EnumDecl.hpp
+++ b/include/AST/Decl/EnumDecl.hpp
@@ -5,8 +5,6 @@
 #include "ASTNode.hpp"
 #include "Types.hpp"
 
-#include <llvm/ADT/SmallVector.h>
-
 namespace glu::ast {
 
 /// @class EnumDecl
@@ -29,7 +27,7 @@ public:
     /// the enum.
     EnumDecl(
         ASTContext &context, SourceLocation location, ASTNode *parent,
-        std::string name, llvm::SmallVector<Case> cases
+        llvm::StringRef name, llvm::ArrayRef<Case> cases
     )
         : DeclBase(NodeKind::EnumDeclKind, location, parent)
         , _self(context.getTypesMemoryArena().create<EnumTy>(
--- a/include/AST/Types/BoolTy.hpp
+++ b/include/AST/Types/BoolTy.hpp
@@ -2,6 +2,7 @@
 #define GLU_AST_TYPES_BOOLTY_HPP
 
 #include "TypeBase.hpp"
+
 namespace glu::types {
 
 /// @brief BoolTy is a class that represents the bool type in the AST.
--- a/include/AST/Types/BoolTy.hpp
+++ b/include/AST/Types/BoolTy.hpp
@@ -3,6 +3,8 @@
 
 #include "TypeBase.hpp"
 
+#include <llvm/ADT/Hashing.h>
+
 namespace glu::types {
 
 /// @brief BoolTy is a class that represents the bool type in the AST.
@@ -21,7 +23,10 @@ public:
 
     /// @brief Method to hash the BoolTy.
     /// @return Returns the hash of the BoolTy.
-    std::size_t hash() const override { return 0; }
+    std::size_t hash() const override
+    {
+        return llvm::hash_value(TypeKind::BoolTyKind);
+    }
 
     /// @brief Method to compare two BoolTy.
     /// @param other The other BoolTy to compare.
--- a/include/AST/Types/BoolTy.hpp
+++ b/include/AST/Types/BoolTy.hpp
@@ -20,14 +20,6 @@ public:
     {
         return type->getKind() == TypeKind::BoolTyKind;
     }
-
-    /// @brief Method to compare two BoolTy.
-    /// @param other The other BoolTy to compare.
-    /// @return Returns `true` if the two BoolTy are equal, `false` otherwise.
-    bool operator==(TypeBase const &other) const override
-    {
-        return other.getKind() == TypeKind::BoolTyKind;
-    }
 };
 
 } // end namespace glu::types
--- a/include/AST/Types/BoolTy.hpp
+++ b/include/AST/Types/BoolTy.hpp
@@ -18,6 +18,18 @@ public:
     {
         return type->getKind() == TypeKind::BoolTyKind;
     }
+
+    /// @brief Method to hash the BoolTy.
+    /// @return Returns the hash of the BoolTy.
+    std::size_t hash() const override { return 0; }
+
+    /// @brief Method to compare two BoolTy.
+    /// @param other The other BoolTy to compare.
+    /// @return Returns `true` if the two BoolTy are equal, `false` otherwise.
+    bool operator==(TypeBase const &other) const override
+    {
+        return other.getKind() == TypeKind::BoolTyKind;
+    }
 };
 
 } // end namespace glu::types
--- a/include/AST/Types/BoolTy.hpp
+++ b/include/AST/Types/BoolTy.hpp
@@ -21,13 +21,6 @@ public:
         return type->getKind() == TypeKind::BoolTyKind;
     }
 
-    /// @brief Method to hash the BoolTy.
-    /// @return Returns the hash of the BoolTy.
-    std::size_t hash() const override
-    {
-        return llvm::hash_value(TypeKind::BoolTyKind);
-    }
-
     /// @brief Method to compare two BoolTy.
     /// @param other The other BoolTy to compare.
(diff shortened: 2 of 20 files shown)


Snowy1803 (changes requested):
Two StructTy aren't necessarily the same struct.

include/AST/Types/BoolTy.hpp
```diff
@@ -18,6 +18,18 @@ class BoolTy : public TypeBase {
     {
         return type->getKind() == TypeKind::BoolTyKind;
     }
+
+    /// @brief Method to hash the BoolTy.
+    /// @return Returns the hash of the BoolTy.
+    std::size_t hash() const override { return 0; }
```
Snowy1803, line 24: At least include the kind in the hashed info

include/Basic/InternedMemoryArena.hpp
```diff
@@ -0,0 +1,70 @@
...
+    static bool isEqual(Base const *lhs, Base const &rhs)
+    {
+        if (lhs == getEmptyKey() || lhs == getTombstoneKey())
+            return false;
+        return *lhs == rhs;
+    }
+};
+
+template <typename Base> class InternedMemoryArena {
+    TypedMemoryArena<Base> _arena;
```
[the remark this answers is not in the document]
LindonAliu: For me, the two have two different functions: one check types, the other internalises the data. I think here it makes more sense to differentiate than to favour inheritance

include/Basic/InternedMemoryArena.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+    static bool isEqual(Base const *lhs, Base const *rhs)
+    {
+        if (lhs == getEmptyKey() || lhs == getTombstoneKey()
+            || rhs == getEmptyKey() || rhs == getTombstoneKey())
+            return lhs == rhs;
+        return *lhs == *rhs;
+    }
+
+    static bool isEqual(Base const *lhs, Base const &rhs)
```
Snowy1803, line 37: are the reference overloads used?

include/Basic/InternedMemoryArena.hpp
```diff
@@ -0,0 +1,70 @@
...
+    {
+        static_assert(
+            std::is_base_of_v<Base, T>, "T must be a subclass of Base"
+        );
+
+        T tmp(std::forward<Args>(args)...);
+
+        auto it = _internedSet.find(&tmp);
+        if (it != _internedSet.end())
+            return static_cast<T *>(*it);
```
Snowy1803, line 60:
```suggestion
            return llvm::cast<T>(*it);
```
more appropriate

include/Basic/InternedMemoryArena.hpp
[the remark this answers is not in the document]
Snowy1803: They have different responsibilities but offer the same interface (a create method). That's when you should use inheritance over composition. Also, now you don't have access to any memory arena functions other than create, which otherwise would be inherited automatically.

include/Basic/TypedMemoryArena.hpp
```diff
@@ -2,7 +2,12 @@
 #define GLU_TYPEDMEMORYARENA_HPP
 
 #include "MemoryArena.hpp"
+#include <llvm/ADT/DenseMapInfo.h>
+#include <llvm/ADT/DenseSet.h>
 #include <type_traits>
+#include <typeindex>
+#include <unordered_map>
+#include <utility>
```
Snowy1803, line 10: none of these imports are used here

test/AST/ASTContext.cpp
```diff
@@ -1,23 +1,88 @@
...
+    ASTContext ctx;
+
+    auto int32Signed
+        = ctx.getTypesMemoryArena().create<IntTy>(IntTy::Signed, 32);
+    auto int32Signed2
+        = ctx.getTypesMemoryArena().create<IntTy>(IntTy::Signed, 32);
+
+    auto int32Unsigned
+        = ctx.getTypesMemoryArena().create<IntTy>(IntTy::Unsigned, 32);
+    // Création d'un IntTy signé 64 bits
```
Snowy1803, line 65: comment

Snowy1803 (commented):
It would be cleaner to use one or two TypeVisitor(s) rather than virtual functions within the type classes.

include/Basic/InternedMemoryArena.hpp
LindonAliu: yes, is used by `find_as()`

include/AST/Types/BoolTy.hpp
```diff
@@ -3,6 +3,8 @@
 
 #include "TypeBase.hpp"
 
+#include <llvm/ADT/Hashing.h>
```
Snowy1803, line 6: useless includes

lib/AST/HashType.cpp
```diff
@@ -0,0 +1,110 @@
...
+class HashVisitor : public TypeVisitor<HashVisitor, std::size_t> {
+public:
+    std::size_t visitTypeBase(TypeBase *type)
+    {
+        return llvm::hash_value(type->getKind());
+    }
+
+    std::size_t visitBoolTy(TypeBase *type) { return visitTypeBase(type); }
+
+    std::size_t visitCharTy(TypeBase *type) { return visitTypeBase(type); }
```
Snowy1803, line 16: these can be omitted — they're the default action

lib/AST/HashType.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+        return hash;
+    };
+
+    std::size_t visitIntTy(IntTy *type)
+    {
+        return llvm::hash_combine(
+            type->getKind(), type->getSignedness(), type->getBitWidth()
+        );
+    }
```
Snowy1803, line 41: what about StructTy, EnumTy, ..?

LindonAliu (commented):
There were memory leaks in tests with types that stored llvm::SmallVector, because llvm::BumpAllocator did not call the destructors of stored objects when it destroyed itself.
I then tried to work around the problem by trying to apply llvm::TrailingObjects on EnumTy with a custom create function that takes the common llvm::BumpAllocator as a parameter in my last commit. This works to avoid memory leaks, the problem is that we are obliged to allocate each time even when we call the ASTContext create function twice with the same arguments because we can no longer construct the type in order to check whether it is already present. What do you think @Snowy1803 ?

include/AST/Types/EnumTy.hpp
```diff
@@ -2,60 +2,98 @@
...
+    ///
+    /// @param allocator The allocator (for example, llvm::BumpPtrAllocator)
+    /// used for allocation.
+    /// @param name The name of the enumeration.
+    /// @param cases The cases to copy into the trailing storage.
+    /// @param loc The definition location.
+    /// @return A pointer to the new EnumTy.
+    static EnumTy *create(
+        llvm::BumpPtrAllocator &allocator, std::string const &name,
+        llvm::SmallVectorImpl<Case> const &cases, SourceLocation loc
```
Snowy1803, line 69: if you're making a copy anyway, take an llvm::ArrayRef
```suggestion
        llvm::ArrayRef<Case> const &cases, SourceLocation loc
```

include/AST/Types/EnumTy.hpp
```diff
@@ -2,60 +2,98 @@
...
+        : TypeBase(TypeKind::EnumTyKind)
+        , _name(std::move(name))
+        , _numCases(numCases)
+        , _definitionLocation(loc)
+    {
+    }
+
 public:
-    /// @brief Constructor for the EnumTy class.
     EnumTy(
```
Snowy1803, line 50: why does this ctor exist?

include/AST/Types/EnumTy.hpp
```diff
@@ -2,60 +2,98 @@
...
+    }
 
-    size_t getCaseCount() const { return _cases.size(); }
-    Case const &getCase(size_t index) const { return _cases[index]; }
+    std::string getName() const { return _name; }
 
-    SourceLocation getDefinitionLocation() const { return _definitionLocation; }
+    unsigned getCaseCount() const { return _numCases; }
 
-    std::optional<size_t> getCaseIndex(llvm::StringRef name) const
```
Snowy1803, line 48: why did you remove this function?

Snowy1803: you can probably make an llvm::ArrayRef from getTrailingObjects() to help you

include/AST/Types/StructTy.hpp
```diff
@@ -11,49 +11,85 @@
 
 namespace glu::types {
 
+struct Field {
+    std::string name;
```
Snowy1803, line 15: std::string allocates, so you'll still end up with a memory leak

Snowy1803: maybe use a Token, as struct field names should be identifiers that are written in the source?

lib/AST/HashType.cpp
```diff
@@ -0,0 +1,204 @@
...
+            return type->getName() == otherName->getName();
+        }
+
+        return false;
+    }
+};
+
+unsigned glu::types::TypeBase::hash(TypeBase const *T) const
+{
+    return HashVisitor().visit(const_cast<TypeBase *>(T));
```
Snowy1803, line 196: why take a parameter instead of using `this`?

include/AST/Types/EnumTy.hpp
LindonAliu: OoOoOps, boulette

include/AST/Types/EnumTy.hpp
LindonAliu: To can construct a simple object to check if the type is already registered in the BumpAllocator from InternedMemoryArena

include/AST/Decl/EnumDecl.hpp
```diff
@@ -28,7 +29,7 @@ class EnumDecl : public DeclBase {
     /// the enum.
     EnumDecl(
         ASTContext &context, SourceLocation location, ASTNode *parent,
-        std::string name, llvm::SmallVector<EnumTy::Case> cases
+        std::string name, llvm::SmallVector<Case> cases
```
Snowy1803, line 32: can use ArrayRef here now, no?

include/AST/Types/EnumTy.hpp
```diff
@@ -2,60 +2,111 @@
...
+        , _name(std::move(name))
+        , _numCases(numCases)
+        , _definitionLocation(loc)
+    {
+    }
+
     EnumTy(
-        std::string name, llvm::SmallVector<Case> cases,
-        SourceLocation definitionLocation
+        std::string const &name, llvm::ArrayRef<Case> const &cases,
```
Snowy1803, line 56: take least specific types. A reference to a reference is useless
```suggestion
        llvm::StringRef name, llvm::ArrayRef<Case> cases,
```

Snowy1803 (approved):

Snowy1803 (approved):

Snowy1803 (approved):

PR #198, opened 2025-02-18
Feature: ASTPrinter Debug

### LINKED ISSUE #11: Create ASTPrinter Debug
Create ASTPrinter lib. Contains a visitor that prints every node in the AST, for debugging.

badermarouan:
Fix #11


include/AST/ASTPrinter.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+    void visitASTNode(ASTNode *node)
+    {
+        if (!node) {
+            out << "Null ASTNode\n";
+            return;
+        }
+        out << toString(node->getKind())
+            << " at loc : " << node->getLocation().getOffset() << "\n";
+    }
```
Snowy1803, line 33: you should have overloads for each kind of node, so you can get and print their children

include/AST/ASTPrinter.hpp
```diff
@@ -0,0 +1,38 @@
...
+    ASTPrinter(llvm::raw_ostream &out = llvm::outs()) : out(out) { }
+
+    void visitASTNode(ASTNode *node)
+    {
+        if (!node) {
+            out << "Null ASTNode\n";
+            return;
+        }
+        out << toString(node->getKind())
+            << " at loc : " << node->getLocation().getOffset() << "\n";
```
Snowy1803, line 32: the location should obviously be (filename, line, column), just like our current token printer

lib/ASTPrinter/ASTStmtPrinter.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+/// @brief Visits an IfStmt node.
+/// @param node The IfStmt node to be visited.
+void ASTPrinter::visitIfStmt(IfStmt *node)
+{
+    beforeVisit(node);
+    out << "\n";
+    visit(node->getCondition());
+    visit(node->getBody());
+    visit(node->getElse());
```
Snowy1803, line 43: visiting the children should be factored in an ASTVisitor subtype: ASTWalker
That's something we'll want to do often, we don't want to write this code each time.

lib/ASTPrinter/ASTStmtPrinter.cpp
```diff
@@ -0,0 +1,111 @@
...
+        out << "\n";
+        afterVisit();
+    }
+}
+
+/// @brief Visits an IfStmt node.
+/// @param node The IfStmt node to be visited.
+void ASTPrinter::visitIfStmt(IfStmt *node)
+{
+    beforeVisit(node);
```
Snowy1803, line 39: before/after callbacks should be factored in the `visit()` implementation of ASTVisitor
We don't want to write before/after in each visit function... and it's also something we can reuse for other visitors. This is already implemented in GILVisitor, you can do the same.

include/AST/ASTPrinter.hpp
```diff
@@ -0,0 +1,84 @@
...
+    default: return out << "Unknown";
+    }
+}
+
+/// @class ASTPrinter
+/// @brief A class that prints the Abstract Syntax Tree (AST) nodes.
+///
+/// This class is derived from ASTVisitor and is used to print the AST nodes
+/// to the specified output stream.
+class ASTPrinter : public ASTVisitor<ASTPrinter> {
```
Snowy1803, line 46: this class should all be in a `.cpp` file — just like the HashVisitor

include/AST/ASTNode.hpp
```diff
@@ -60,6 +61,10 @@ class ASTNode {
     /// @brief Get the location of the current node.
     /// @return The location of the current node.
     SourceLocation getLocation() const { return _nodeLocation; }
+
+    void debugPrint(
+        glu::SourceManager *srcManager, llvm::raw_ostream &out = llvm::outs()
```
Snowy1803, line 66: ASTNodes should probably have access to the source manager somehow — probably through the root node

include/AST/Stmt/IfStmt.hpp
```diff
@@ -36,8 +36,13 @@ class IfStmt : public StmtBase {
...
     {
-        assert(condition && "Condition cannot be null");
-        assert(body && "Body cannot be null");
+        assert(_condition && "Condition cannot be null.");
+        assert(_body && "Body cannot be null.");
+        assert(
+            (!elseBranch)
+            || (elseBranch->getParent() == nullptr)
+                && "Else branch already has a parent."
+        );
```
Snowy1803, line 45: this assert is pretty useless — the parser would never put the same node at different places

include/Basic/SourceManager.hpp
```diff
@@ -169,17 +166,17 @@ class SourceManager {
...
     }
 
-    void reset()
-    {
-        _fileLocEntries.clear();
-        _nextOffset = 0;
-        _mainFile = FileID(0);
-    }
+    /// @brief Reset the SourceManager to its initial state.
+    void reset();
```
Snowy1803, line 179: I still don't see why this function exists — it shouldn't

If ever someone needs to reset a source manager (I can't think of why that would ever be the case), it would just be:
```
SourceManager newSM;
std::swap(oldSM, newSM);
```

lib/ASTPrinter/ASTPrinter.cpp
```diff
@@ -1,5 +1,315 @@
...
+    }
+    kindStr.consume_back("Ty");
+    return out << kindStr;
+}
+
+llvm::raw_ostream &
+operator<<(llvm::raw_ostream &out, glu::types::Field const &c)
+{
+    return out << c.name << " = " << c.type->getKind();
+}
```
Snowy1803, line 43: why's the structure one here and the enum one in the enum file?

lib/ASTPrinter/ASTPrinter.cpp
```diff
@@ -1,5 +1,315 @@
...
+        : _srcManager(srcManager), out(out)
+    {
+    }
+
+    /// @brief Visits an AST node.
+    /// @param node The AST node to be visited.
+    void visitASTNode(ASTNode *node)
+    {
+        if (!node)
+            out << "Null ASTNode\n";
```
Snowy1803, line 105: that's impossible — the ASTVisitor would crash before this function is called

lib/ASTPrinter/ASTPrinter.cpp
```diff
@@ -1,5 +1,315 @@
...
+#include "Basic/TokenKind.def"
+    default: return out << "Unknown";
+    }
+}
+
+class ASTPrinter : public ASTWalker<ASTPrinter, TraversalOrder::PreOrder> {
+    SourceManager *_srcManager; ///< The source manager.
+    llvm::raw_ostream &out; ///< The output stream to print the AST nodes.
+    size_t _indent = 0; ///< The current indentation level.
+    unsigned _printDetails = 0; ///< Whether to print details of the nodes.
```
Snowy1803, line 59: what is that? it makes no sense

lib/ASTPrinter/ASTPrinter.cpp
```diff
@@ -1,5 +1,315 @@
...
+
+    /// @brief Visits a VarLetDecl node.
+    /// @param node The VarLetDecl node to be visited.
+    void visitVarLetDecl(VarLetDecl *node)
+    {
+        out.indent(_indent);
+        out << "-->Name: " << node->getName() << '\n';
+        out.indent(_indent);
+        out << "-->Type: " << node->getType()->getKind() << '\n';
+    }
```
Snowy1803, line 222: VarLetDecl isn't a NodeKind that exists, so this can never be called

Snowy1803: Did you mean LetDecl?

lib/ASTPrinter/ASTPrinter.cpp
```diff
@@ -1,5 +1,315 @@
...
+    void visitStructMemberExpr(StructMemberExpr *node)
+    {
+        out.indent(_indent - 4);
+        out << "-->Member: " << node->getMemberName() << " from struct:\n";
+    }
+
+    void visitUnaryOpExpr(UnaryOpExpr *node)
+    {
+        out.indent(_indent - 4);
+        out << "-->" << node->getOperator() << " Unary Operation with:\n";
```
Snowy1803, line 305: everywhere, you don't need to write the kind again, you already printed that it was a UnaryOp in beforeVisit, and you don't need to say "with:" because we know it's a tree and there's already indentation and arrows that are clear enough. we don't need more unnecessary text.

include/Basic/SourceManager.hpp
badermarouan: I used it for the units test

Snowy1803 (approved):

PR #201, opened 2025-02-23
AST: add `ImportDecl`

### LINKED ISSUE #56: Implement `glu::ast::ImportDecl` class

LindonAliu:
fix #56

### DIFF
--- a/include/AST/Decl/ImportDecl.hpp
+++ b/include/AST/Decl/ImportDecl.hpp
@@ -1,0 +1,52 @@
+#ifndef GLU_AST_DECL_IMPORTDECL_HPP
+#define GLU_AST_DECL_IMPORTDECL_HPP
+
+#include "ASTNode.hpp"
+#include "Types.hpp"
+
+namespace glu::ast {
+
+/// @struct ImportPath
+/// @brief Represents the import path as components.
+/// For example, for "std::io::{println}", components will contain ["std", "io"]
+/// and selectors will contain ["println"].
+struct ImportPath {
+    std::vector<llvm::StringRef> components;
+    std::vector<llvm::StringRef> selectors;
+};
+
+/// @class ImportDecl
+/// @brief Represents an import declaration in the AST.
+///
+/// This class inherits from DeclBase and encapsulates the details of an import
+/// declaration.
+class ImportDecl : public DeclBase {
+    ImportPath _importPath;
+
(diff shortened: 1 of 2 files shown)


include/AST/Decl/ImportDecl.hpp
```diff
@@ -0,0 +1,52 @@
...
+#include "ASTNode.hpp"
+#include "Types.hpp"
+
+namespace glu::ast {
+
+/// @struct ImportPath
+/// @brief Represents the import path as components.
+/// For example, for "std::io::{println}", components will contain ["std", "io"]
+/// and selectors will contain ["println"].
+struct ImportPath {
```
Snowy1803, line 13: Please document how you encode:
```
import std;
import std::io::println;
import std::*;
```

include/AST/Decl/ImportDecl.hpp
```diff
@@ -0,0 +1,72 @@
...
+///    - components: ["std"]
+///    - selectors: []
+///
+/// - For "std::io":
+///    - components: ["std", "io"]
+///    - selectors: []
+///
+/// - For "std::io::println":
+///    - components: ["std", "io"]
+///    - selectors: ["println"]
```
Snowy1803, line 32: How can you grammatically know if something is a module or a function? I think the last element should always be in selectors, which should never be empty

LindonAliu: Imo, components should be separated from selectors in an ungrammatical way, as if you were choosing what you wanted from the last component with selector
 It's the same thing, but grammatically you don't really have the choice of doing as you say

include/AST/Decl/ImportDecl.hpp
```diff
@@ -0,0 +1,72 @@
...
+/// - For "std::io":
+///    - components: ["std"]
+///    - selectors: ["io"]
+///
+/// - For "std::io::println":
+///    - components: ["std", "io"]
+///    - selectors: ["println"]
+struct ImportPath {
+    std::vector<llvm::StringRef> components;
+    std::vector<llvm::StringRef> selectors;
```
Snowy1803, line 35: Note that this std vector is still allocated on the heap and will leak

Snowy1803 (approved):

Snowy1803 (approved):

PR #281, opened 2025-03-16
Feature: Add DebugInst

### LINKED ISSUE #32: Create `glu::gil::DebugInst` class
Also, add source location to all instructions

LindonAliu:
fix #32


include/GIL/Instructions/DebugInst.hpp
```diff
@@ -0,0 +1,61 @@
...
+#include "InstBase.hpp"
+
+namespace glu::gil {
+
+class DebugInst : public InstBase {
+public:
+    enum class DebugBindingType {
+        Let,
+        var,
+        Arg,
```
Snowy1803, line 13: use a consistent case? SCREAMING_SNAKE_CASE I'd say.

Snowy1803: also I'd put this enum out of the DebugInst class, glu::gil::DebugInst::DebugBindingType::LET is a bit long

include/GIL/Instructions/DebugInst.hpp
```diff
@@ -0,0 +1,61 @@
...
+    llvm::StringRef getName() const { return _name; }
+    Value getValue() const { return _value; }
+    DebugBindingType getBindingType() const { return _bindingType; }
+    SourceLocation getLocation() const { return InstBase::getLocation(); }
+
+    size_t getResultCount() const override { return 0; }
+    Type getResultType([[maybe_unused]] size_t index) const override
+    {
+        assert(false && "DebugInst has no results");
+        return Type();
```
Snowy1803, line 43: use llvm::unreachable instead of assert(false) and a fake return

include/GIL/Instructions/DebugInst.hpp
LindonAliu: woooooooooooooooops

Snowy1803 (commented):
Can you update `GILPrinter` to print the source location of each instruction (via `, loc "thing.glu":1:1` if not null), and show the metadata in `DebugInst`?

lib/GIL/GILPrinter.cpp
```diff
@@ -147,4 +147,30 @@ void GILPrinter::printLabel(BasicBlock *bb)
     }
 }
 
+void GILPrinter::printSourceLocation(SourceLocation loc)
+{
+    if (!loc.isValid() || !sm.isInMainFile(loc))
```
Snowy1803, line 152: Why only print if in main file? We can have other files

lib/GIL/GILPrinter.cpp
```diff
@@ -147,4 +147,30 @@ void GILPrinter::printLabel(BasicBlock *bb)
...
+{
+    out << inst->getInstName();
+
+    size_t operands = inst->getOperandCount();
+    for (size_t i = 0; i < operands; ++i) {
+        if (i != 0)
+            out << ",";
+        out << " ";
+        printOperand(inst->getOperand(i));
+    }
```
Snowy1803, line 170: Maybe factor this into a printOperands instead of copying the code?

lib/GIL/GILPrinter.cpp
```diff
@@ -82,13 +82,7 @@ void GILPrinter::visitInstBase(InstBase *inst)
...
     }
     out << inst->getInstName();
-    size_t operands = inst->getOperandCount();
-    for (size_t i = 0; i < operands; ++i) {
-        if (i != 0)
-            out << ",";
-        out << " ";
-        printOperand(inst->getOperand(i));
-    }
+    printOperands(inst);
```
Snowy1803, line 85: other instructions can have locations as well, print it?

Snowy1803 (approved):

PR #288, opened 2025-03-23
 Feature: Create structure for contraints

### LINKED ISSUE #267: Create structure for Contraints

badermarouan:
Fix #267


include/Sema/Constraints.hpp
```diff
@@ -0,0 +1,466 @@
...
+    ArgumentConversion, ///< Conversion for function arguments.
+    OperatorArgumentConversion, ///< Conversion for operator arguments.
+    CheckedCast, ///< Checked cast from first to second type.
+    BindOverload, ///< Binds to a specific overload.
+    ValueMember, ///< First type has a value member of second type.
+    UnresolvedValueMember, ///< Like ValueMember, but implicit base.
+    Defaultable, ///< First type can default to second.
+    Disjunction, ///< One or more constraints must hold.
+    Conjunction, ///< All constraints must hold.
+    FallbackType, ///< Used when no contextual type exists.
```
Snowy1803, line 32: difference with Defaultable?

include/Sema/Constraints.hpp
```diff
@@ -0,0 +1,466 @@
...
+    ExplicitGenericArguments, ///< Explicit generic args for overload.
+    LValueObject, ///< First is l-value, second is object type.
+
+    /// Represents an AST node contained in a body of a
+    /// function/closure.
+    /// It only has an AST node to generate constraints and
+    /// infer the type for.
+    SyntacticElement
```
Snowy1803, line 40: what

include/Sema/Constraints.hpp
```diff
@@ -0,0 +1,466 @@
...
+    DeepEquality, ///< Deep structural equality.
+    ArrayToPointer, ///< Array to pointer conversion.
+    StringToPointer, ///< String to pointer conversion.
+    PointerToPointer ///< Pointer to pointer conversion.
+};
+} // namespace glu::sema
+
+namespace glu::types {
+using Ty = glu::types::TypeBase *;
+}
```
Snowy1803, line 70: this should just be in TypeBase.hpp?

include/Sema/Constraints.hpp
```diff
@@ -0,0 +1,466 @@
...
+    using TrailingArgs = llvm::TrailingObjects<
+        Constraint, glu::types::TypeVariableTy *, glu::ast::FunctionDecl *>;
+
+    // Make the TrailingObjects base class a friend to allow proper access
+    friend TrailingArgs;
+
+    ConstraintKind _kind; ///< Kind of constraint.
+
+    ConversionRestrictionKind
+        _restriction : 8; ///< Optional conversion restriction.
```
Snowy1803, line 94: why a bitfield if alone? just add a char type to the enum

include/Sema/Constraints.hpp
```diff
@@ -0,0 +1,466 @@
...
+    unsigned _rememberChoice : 1; ///< Should solver record disjunction choice.
+    unsigned _isFavored : 1; ///< Preferred constraint for disjunctions.
+    unsigned _isDiscarded : 1; ///< Whether the result is unused.
+    unsigned
+        _hasDeclContext : 1; ///< Whether the constraint has a decl context.
+
+    union {
+        struct {
+            glu::types::Ty First; ///< First type involved.
+            glu::types::Ty Second; ///< Second type involved.
```
Snowy1803, line 110: naming convention

include/Sema/Constraints.hpp
```diff
@@ -0,0 +1,466 @@
...
+        struct {
+            glu::types::Ty First; ///< Base type.
+            glu::types::Ty Second; ///< Member type.
+            glu::ast::StructMemberExpr *structMember; ///< Member node.
+        } _member;
+
+        llvm::ArrayRef<Constraint *> _nested; ///< Nested constraints.
+
+        struct {
+            glu::types::Ty First; ///< Overload target type.
```
Snowy1803, line 122: overload should be a DeclBase *

include/Sema/Constraints.hpp
```diff
@@ -0,0 +1,466 @@
...
+    size_t numTrailingObjects(TrailingObjects::OverloadToken<T>) const
+    {
+        if constexpr (std::is_same_v<T, glu::types::TypeVariableTy *>) {
+            return _numTypeVariables;
+        } else if constexpr (std::is_same_v<T, glu::ast::FunctionDecl *>) {
+            return (_kind == ConstraintKind::BindOverload) ? 1 : 0;
+        } else {
+            return 0;
+        }
+    }
```
Snowy1803, line 238: you're supposed to write different function overloads... not a templated one with an if?

include/Sema/Constraints.hpp
```diff
@@ -0,0 +1,466 @@
...
+            _kind == ConstraintKind::Disjunction
+            || _kind == ConstraintKind::Conjunction
+        );
+        return _nested;
+    }
+
+    glu::ast::FunctionDecl *getOverloadChoice() const
+    {
+        assert(_kind == ConstraintKind::BindOverload);
+        return *getTrailingObjects<glu::ast::FunctionDecl *>();
```
Snowy1803, line 424: why would you store it in trailing objects if there's just one?

lib/Sema/Constraints.cpp
```diff
@@ -0,0 +1,464 @@
...
+        for (auto nested : constraint->getNestedConstraints())
+            gatherReferencedTypeVars(nested, typeVars);
+        return;
+
+    case ConstraintKind::Conjunction:
+        typeVars.insert(
+            constraint->getTypeVariables().begin(),
+            constraint->getTypeVariables().end()
+        );
+        return;
```
Snowy1803, line 192: why the difference?

lib/Sema/Constraints.cpp
```diff
@@ -0,0 +1,464 @@
...
+    case ConstraintKind::Defaultable:
+    case ConstraintKind::FallbackType:
+    case ConstraintKind::ExplicitGenericArguments:
+    case ConstraintKind::LValueObject:
+        typeVars.insert(
+            llvm::cast<glu::types::TypeVariableTy>(constraint->getFirstType())
+        );
+        typeVars.insert(
+            llvm::cast<glu::types::TypeVariableTy>(constraint->getSecondType())
+        );
```
Snowy1803, line 211: llvm::cast will crash if the cast fails... and there's no point in constraints if everything's a type variable, so of course it will fail

lib/Sema/Constraints.cpp
```diff
@@ -0,0 +1,464 @@
...
+    , _locator(locator)
+{
+    std::copy(
+        typeVars.begin(), typeVars.end(), getTypeVariablesBuffer().begin()
+    );
+    *getTrailingObjects<glu::ast::FunctionDecl *>() = choice;
+}
+
+/// Recursively gather the set of type variables referenced by this constraint.
+static void gatherReferencedTypeVars(
```
Snowy1803, line 176: what's the point of this??

lib/Sema/Constraints.cpp
```diff
@@ -0,0 +1,464 @@
...
+)
+{
+    llvm::SmallPtrSet<glu::types::TypeVariableTy *, 4> typeVars;
+
+    assert(first && "First type is Null");
+    assert(second && "Second type is Null");
+    assert(locator && "Locator is Null");
+
+    if (first->getKind() == glu::types::TypeKind::TypeVariableTyKind)
+        typeVars.insert(llvm::cast<glu::types::TypeVariableTy>(first));
```
Snowy1803, line 252: never check the kind directly... use llvm::isa or
```suggestion
    if (auto *ty = llvm::dyn_cast<glu::types::TypeVariableTy>(first))
        typeVars.insert(ty);
```

include/Sema/Constraints.hpp
badermarouan: to avoid allocating space for this in constraints that won't need it

lib/Sema/Constraints.cpp
badermarouan: It's a function that can gather every variables used for any type of constraint even disjunction or conjunction by collecting them recursively

lib/Sema/Constraints.cpp
badermarouan: Because in a conjunction all the variables are already stored in a smallPtrSet due to the fact that this constraint represents a logical "and" between its constraints
Whereas a disjunction will store their variables to differents storage and so will need to collect those variables in these storages recursively due to a disjunction being a logical "or" between its constraints

badermarouan (commented):
Add a specific create function for each constraint

include/Sema/Constraints.hpp
Snowy1803: Should be stored in the union

include/Sema/Constraints.hpp
```diff
@@ -0,0 +1,465 @@
...
+    /// @param first The first type in the constraint.
+    /// @param second The second type in the constraint.
+    /// @param locator The AST node that triggered this constraint.
+    /// @param extraTypeVars Additional type variables to include.
+    /// @return A newly created constraint.
+    static Constraint *create(
+        llvm::BumpPtrAllocator &allocator, ConstraintKind kind,
+        glu::types::Ty first, glu::types::Ty second, glu::ast::ASTNode *locator,
+        llvm::ArrayRef<glu::types::TypeVariableTy *> extraTypeVars = {}
+    );
```
Snowy1803, line 250: One create function for each kind would be clearer

lib/Sema/Constraints.cpp
Snowy1803: they should all contain the same type variables? a type variable shouldn't be left unresolved for one of the cases in a disjunction

badermarouan: In a Disjunction some constraints cannot be resolved and so needs different type variables, for example in case of a BindOverload:
if we call :
`foo(42)`
and we have:
`func foo(Int)
func foo(String)`
we have two constraints in this disjunction:
> Disjunction {
>   BindOverload to func foo(Int)
>   BindOverload to func foo(String)
> }
the first one will be resolved while the second one will fail.
This doesn't apply to a conjunction because if one of the constraints fails the whole conjunction fails which is why we can use the same typeVariables

lib/Sema/Constraints.cpp
```diff
@@ -0,0 +1,570 @@
...
+            );
+        } else if (unwrappedAny) {
+            // Since we unwrapped constraints before, add this constraint.
+            unwrapped.push_back(constraint);
+        }
+        ++index;
+    }
+
+    // If we unwrapped anything, our list of constraints is the unwrapped list.
+    if (unwrappedAny)
```
Snowy1803, line 523: why complicate the code with 2 variables, just use one

lib/Sema/Constraints.cpp
```diff
@@ -0,0 +1,570 @@
...
+    if (constraints.size() == 1) {
+        assert(!rememberChoice && "simplified an important disjunction?");
+        return constraints.front();
+    }
+
+    // Create the disjunction constraint.
+    auto size = totalSizeToAlloc<glu::types::TypeVariableTy *>(typeVars.size());
+    void *mem = allocator.Allocate(size, alignof(Constraint));
+    auto disjunction = new (mem)
+        Constraint(ConstraintKind::Disjunction, constraints, locator, typeVars);
```
Snowy1803, line 538: you are storing an ArrayRef which is a reference to a stack-allocated array which will be deallocated at the end of the function. this will ofc crash. allocate the `constraints` array within the `allocator` so it gets deallocated at the same time as the rest.

include/Sema/Constraints.hpp
```diff
@@ -0,0 +1,569 @@
...
+///
+enum class ConversionRestrictionKind : char {
+    DeepEquality, ///< Deep structural equality.
+    ArrayToPointer, ///< Array to pointer conversion.
+    StringToPointer, ///< String to pointer conversion.
+    PointerToPointer ///< Pointer to pointer conversion.
+};
+} // namespace glu::sema
+
+namespace glu::sema {
```
Snowy1803, line 61: why close and reopen the same ns

include/Sema/Constraints.hpp
```diff
@@ -0,0 +1,569 @@
+#ifndef GLU_SEMA_CONSTRAINTS_HPP
```
Snowy1803, line 1: why `Constraints` if the class name is `Constraint`

lib/Sema/Constraints.cpp
```diff
@@ -0,0 +1,381 @@
...
+    );
+
+    // Create the disjunction constraint.
+    auto disjunction = new (allocator)
+        Constraint(ConstraintKind::Disjunction, nested, locator);
+    disjunction->_rememberChoice = (bool) rememberChoice;
+    return disjunction;
+}
+
+Constraint *Constraint::createMemberOrOuterDisjunction(
```
Snowy1803, line 354: I have no idea what this is for

lib/Sema/Constraints.cpp
```diff
@@ -0,0 +1,354 @@
...
+)
+{
+    assert(first && "First type is Null");
+    assert(second && "Second type is Null");
+    assert(locator && "Locator is Null");
+
+    return ::new (allocator) Constraint(kind, first, second, locator);
+}
+
+Constraint *Constraint::createBind(
```
Snowy1803, line 135: there is no reason to outline all these one-liners

include/Sema/Constraints.hpp
```diff
@@ -0,0 +1,624 @@
...
+        glu::types::Ty first, glu::types::Ty second, glu::ast::ASTNode *locator
+    );
+
+    /// @brief Create a bind constraint between two types.
+    /// @param allocator The allocator for memory allocation.
+    /// @param first The first type in the bind.
+    /// @param second The second type in the bind.
+    /// @param locator The AST node that triggered this constraint.
+    /// @return A newly created bind constraint.
+    inline static Constraint *createBind(
```
Snowy1803, line 206: don't explicitly write the `inline` keyword... I can see that it's in the class, I don't need extra noise
> A function defined entirely inside a class/struct/union definition, whether it's a member function or a non-member friend function, is implicitly an inline function.

[inline specifier](https://en.cppreference.com/w/cpp/language/inline)

Snowy1803: your other inline function don't have the keyword either

Snowy1803 (approved):

PR #296, opened 2025-03-26
Feature: GILGen base, Break, Continue

### LINKED ISSUE #268: Generate GIL for FunctionDecl
Create gil::Function from ast::FunctionDecl

### LINKED ISSUE #269: Generate GIL for Break/ContinueStmt
Keep scope information to be able to break/continue to some basic block.

Snowy1803:
Closes #268 
Closes #269


include/GILGen/Context.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+namespace glu::gilgen {
+
+/// @brief The context/builder for the GIL code generation.
+class Context {
+    gil::Function *_function;
+    gil::BasicBlock *_currentBB;
+    gil::InstBase *_insertBefore = nullptr; // Insert at end of block by default
+    ast::FunctionDecl *_functionDecl;
+    llvm::BumpPtrAllocator &_arena;
```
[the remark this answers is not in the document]
LindonAliu: or make Context inherit from that with custom create ?

lib/GILGen/Context.cpp
```diff
@@ -0,0 +1,16 @@
...
+
+glu::gilgen::Context::Context(
+    ast::FunctionDecl *decl, llvm::BumpPtrAllocator &arena
+)
+    : _functionDecl(decl), _arena(arena)
+{
+    _function
+        = new (_arena) gil::Function(decl->getName().str(), decl->getType());
+    _currentBB = new (_arena)
+        gil::BasicBlock("entry"); // TODO: args are the function's arg
```
LindonAliu, line 14: maybe a static create function is better than call the new operator (to follow what we're doing for AST)

lib/GILGen/GILGenStmt.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include "ASTVisitor.hpp"
+#include "GILGen.hpp"
+
+namespace glu::gilgen {
+
+using namespace glu::ast;
+
+struct GILGenStmt : public ASTVisitor<GILGenStmt, void> {
+    Context ctx;
+    std::vector<Scope> scopes;
```
LindonAliu, line 16: llvm::something maybe more optimized ?

test/GILGen/GILGenStmt.cpp
```diff
@@ -0,0 +1,39 @@
...
+    std::unique_ptr<llvm::MemoryBuffer> buf(          \
+        llvm::MemoryBuffer::getMemBufferCopy(str)     \
+    );                                                \
+    glu::Scanner scanner(buf.get());                  \
+    glu::ast::ASTContext context;                     \
+    glu::SourceManager sm;                            \
+    sm.loadBuffer(std::move(buf), "main.glu");        \
+    glu::Parser parser(scanner, context, sm /*, 1*/); \
+    EXPECT_TRUE(parser.parse());                      \
+    auto module = parser.getAST();
```
LindonAliu, line 21: maybe better to do functionnal tests

lib/GILGen/Context.cpp
Snowy1803: It's simple this way

test/GILGen/GILGenStmt.cpp
Snowy1803: Probably yes

lib/GILGen/Context.cpp
LindonAliu: but more safe if its wrapped, as u want

Snowy1803: it is not safer to have a wrapper doing the exact same thing. there is no typing to add, as Function, BasicBlock and InstBase don't have a common ancestor. Function and BasicBlock also don't have TrailingObjects. InstBase may, but it's abstracted away behind Context anyway. Context is the wrapper.

test/GILGen/GILGenStmt.cpp
Snowy1803: This should be done in a separate PR, once the driver is done.

LindonAliu (approved):

PR #307, opened 2025-03-26
 Feature: add a first implementation of Driver

### LINKED ISSUE #18: Create Driver

LindonAliu:
fix #18 

i've add -fno-rtti due to this error : https://discourse.llvm.org/t/undefined-reference-to-typeinfo-for-llvm-genericoptionvalue/71526


lib/Parser/Parser.yy
```diff
@@ -919,6 +880,7 @@ primary_expression:
 
 argument_list_opt:
       %empty
+        { $$ = llvm::SmallVector<ExprBase*>(); }
```
Snowy1803, line 883: ? the stack slot is already default initialized

test/functionnal/assets/enum.glu
```diff
@@ -0,0 +1 @@
+enum E: Int { A, B = 2 }
```
Snowy1803, line 1: functional with one n (directory name)

LindonAliu: wooooooooops

lib/Parser/Parser.yy
LindonAliu: I've added this line to delete this warning:
image

Snowy1803: You can just {}, I think there’s no need to explicitly re-initialize it

LindonAliu: It works, but is having fewer lines at the expense of clarity what we want?

tools/gluc/sources/main.cpp
```diff
@@ -1,36 +1,50 @@
...
-    if (!fileID) {
-        llvm::errs() << fileID.getError().message() << "\n";
+int main(int argc, char **argv)
+{
+    cl::ParseCommandLineOptions(argc, argv);
+
+    if (InputFilenames.empty()) {
+        errs() << "Usage: " << argv[0] << " <input files...>\n";
         return 1;
     }
```
Snowy1803, line 25: this can't happen, InputFilenames is set as `OneOrMore` so the Parse above fails and exits first

tools/gluc/sources/main.cpp
```diff
@@ -1,36 +1,50 @@
...
-        return 1;
-    }
+#include "llvm/Support/CommandLine.h"
 
-    glu::SourceManager sourceManager;
+using namespace llvm;
+
+static cl::opt<bool> PrintAST(
+    "print-ast", cl::desc("Print the AST after parsing"), cl::init(false)
+);
```
Snowy1803, line 12: add options for PrintTokens and outputting GIL ?

lib/Parser/Parser.yy
```diff
@@ -57,8 +59,11 @@
 
 %code {
     // Redefine yylex to call our scanner and return a symbol
-    static glu::BisonParser::symbol_type yylex(glu::Scanner& scanner) {
+    static glu::BisonParser::symbol_type yylex(glu::Scanner& scanner, bool printTokens) {
         glu::Token tok = scanner.nextToken();
+        if (printTokens) {
+            llvm::outs() << tok << "\n";
+        }
```
Snowy1803, line 66: the Parser should certainly not be concerned with driver output options — that is a wrong separation of concerns
also why print in llvm::outs() and not a customizable file? it should all be handled in the driver.

Snowy1803 (approved):

lib/ASTPrinter/ASTPrinter.cpp
```diff
@@ -48,16 +48,6 @@ operator<<(llvm::raw_ostream &out, glu::types::Field const &c)
...
-llvm::raw_ostream &operator<<(llvm::raw_ostream &out, glu::Token const &token)
-{
-    switch (token.getKind()) {
-#define TOKEN(Name)                                      \
-    case glu::TokenKind::Name##Tok: return out << #Name;
-#include "Basic/TokenKind.def"
-    default: return out << "Unknown";
-    }
-}
-
```
badermarouan, line 60: There is also the same function in ASTPrinter you could use the new one you defined in the ASTPrinter too?

PR #310, opened 2025-03-31
feat: basic of enum variant

yunsecode:
Closes: #46


.gitignore
```diff
@@ -6,6 +6,7 @@
 
 # Cmake build directory
 build/
+buildTest/
```
Snowy1803, line 9: ?

include/GIL/Instructions/EnumVariantInst.hpp
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    static bool classof(InstBase const *inst)
+    {
+        return inst->getKind() == InstKind::EnumVariantInstKind;
+    }
+
+    Operand getOperand(size_t index) const override
+    {
+        assert(false && "EnumVariantInst has no operands");
```
Snowy1803, line 35: It has one operand, the member « return member ; »

include/GIL/Instructions/EnumVariantInst.hpp
```diff
@@ -0,0 +1,50 @@
...
+    Member member; ///< The enum variant represented as a Member.
+
+public:
+    /// @brief Constructs an EnumVariantInst from a Member.
+    ///
+    /// @param member The enum variant to represent.
+    EnumVariantInst(Member member)
+        : ConstantInst(InstKind::EnumVariantInstKind), member(member)
+    {
+        // return member.getType();
```
Snowy1803, line 25: You can assert that the member is an enum and not a struct

include/GIL/Instructions/EnumVariantInst.hpp
```diff
@@ -0,0 +1,53 @@
...
+    {
+        return inst->getKind() == InstKind::EnumVariantInstKind;
+    }
+
+    Operand getOperand(size_t index) const override
+    {
+        if (index == 0) {
+            return Operand(member);
+        }
+        llvm_unreachable("Invalid operand index");
```
Snowy1803, line 38:
```suggestion
        assert(index == 0 && "Invalid operand index");
        return member;
```
it's simpler, clearer, and more predictable code gen.
in debug mode it's equivalent (it will crash for invalid input)
in release mode:
- with assert, it will simply return member no matter the given index, as the assert is just removed. 
- with llvm_unreachable, it will invoke UB if the index is not 0 (it will probably just remove the if and do the same as above)

include/GIL/Instructions/EnumVariantInst.hpp
Snowy1803: add that assert please

Snowy1803 (approved):

Issue #303, opened 2025-03-26
Update GIL data structures to not allocate


Snowy1803:
std::string -> llvm::StringRef
std::vector -> TrailingObjects
etc

- [x] gil::Function (name)
- [x] gil::Module (name)
- [x] gil::BasicBlock (label, args)
- [x] CallInst (args)
- [x] IntegerLiteralInst (what to do with APInt??)
- [x] StringLiteralInst

PR #316, opened 2025-04-07
feat: not allocate string in gil function.hpp

yunsecode:
Closes: #303

### DIFF
--- a/include/GIL/Function.hpp
+++ b/include/GIL/Function.hpp
@@ -61,12 +61,12 @@ private:
                          // when added
 
     BBListType _basicBlocks;
-    std::string _name;
+    llvm::StringRef _name;
     glu::types::FunctionTy *_type;
 
 public:
-    Function(std::string name, glu::types::FunctionTy *type)
-        : _name(std::move(name)), _type(type)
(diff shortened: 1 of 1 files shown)


lib/GILGen/Context.cpp
```diff
@@ -10,7 +10,6 @@ glu::gilgen::Context::Context(
 {
     _function
         = new (_arena) gil::Function(decl->getName().str(), decl->getType());
-    _currentBB = new (_arena)
-        gil::BasicBlock("entry"); // TODO: args are the function's arg
+    _currentBB = gil::BasicBlock::create(_arena, "entry", {});
```
Snowy1803, line 13: either keep the TODO or pass the args — it shouldn't be empty

test/GIL/GILPrinter.cpp
```diff
@@ -51,8 +53,10 @@ TEST_F(GILPrinterTest, SimpleFunction)
 
 TEST_F(GILPrinterTest, FunctionWithArguments)
 {
+    llvm::BumpPtrAllocator alloc;
```
Snowy1803, line 56: you can just put that in the fixture

Snowy1803 (approved):

PR #339, opened 2025-04-20
 Feature: Global scope building

### LINKED ISSUE #338: Create global scope
A visitor that collects declarations in the global scope: structs, enums, types, functions, operators, etc...

Snowy1803:
Closes #338


include/Sema/ScopeTable.hpp
```diff
@@ -0,0 +1,166 @@
...
+    /// It is the scope that contains all the types and functions declared
+    /// in the module.
+    bool isGlobalScope() const { return _parent == nullptr; }
+
+    /// @brief Returns true if this scope is a function scope.
+    /// A function scope is a scope that contains a function declaration.
+    bool isFunctionScope() const
+    {
+        return llvm::isa<ast::FunctionDecl>(_node->getParent());
+    }
```
badermarouan, line 69: Will this crash if getParent returns null (in case the node is a moduleDecl)

include/Sema/ScopeTable.hpp
```diff
@@ -0,0 +1,166 @@
...
+
+    /// @brief Returns the function declaration this scope belongs to,
+    /// or nullptr if this scope is the global scope.
+    ast::FunctionDecl *getFunctionDecl()
+    {
+        if (isGlobalScope())
+            return nullptr;
+        if (isFunctionScope())
+            return llvm::cast<ast::FunctionDecl>(_node->getParent());
+        return _parent->getFunctionDecl();
```
badermarouan, line 101: Will the _parent always be a function even in case of imbricated ifs ?

include/Sema/ScopeTable.hpp
```diff
@@ -0,0 +1,166 @@
...
+    /// This is used to resolve overloaded functions and variables.
+    /// Note that if there are multiple overloads, in different scopes,
+    /// the ones in the closest scope are returned.
+    ScopeItem *lookupItem(llvm::StringRef name)
+    {
+        auto it = _items.find(name);
+        if (it != _items.end())
+            return &it->second;
+        if (_parent)
+            return _parent->lookupItem(name);
```
badermarouan, line 116: Does this also look for included modules Items? Or this will be checked in top level logic using all the global scopes?

include/Sema/ScopeTable.hpp
```diff
@@ -0,0 +1,166 @@
...
+    /// @return A pointer to the DeclBase if found, or nullptr if not found.
+    ast::TypeDecl *lookupType(llvm::StringRef name)
+    {
+        // Note: only the global scope should have types, but we check all
+        // scopes because why not
+        auto it = _types.find(name);
+        if (it != _types.end())
+            return it->second;
+        if (_parent)
+            return _parent->lookupType(name);
```
badermarouan, line 131: Same does this also look for included modules Types?

include/Sema/ScopeTable.hpp
Snowy1803: In case of imbricated unnamed scopes (if, for, etc), there will always be a function in the ancestors. The function is recursive so it will always find a function. We don’t have other kinds of scopes (classes, methods, etc)

include/Sema/ScopeTable.hpp
Snowy1803: Import is not implemented yet, this will be done later, after the beta.

Snowy1803: When it will be implemented, I expect specified imports `import file::helloFromFile;` to be added to the global scope, but module imports`import file;` will require a namespace and so the namespace will be its own item which is a subscope corresponding to an import. Those subscopes will themselves be “global scopes” for their own module.

include/Sema/ScopeTable.hpp
Snowy1803: yes, fixed

badermarouan (approved):

LindonAliu (approved):

PR #347, opened 2025-04-23
Feature: Diagnostic

### LINKED ISSUE #340: Create Diagnostic class (Basic)

LindonAliu:
Closes #340

### DIFF
--- a/include/Basic/Diagnostic.hpp
+++ b/include/Basic/Diagnostic.hpp
@@ -20,7 +20,7 @@ enum class DiagnosticSeverity {
 };
 
 /// @brief A diagnostic message with location and severity information.
-class DiagnosticMessage {
+class Diagnostic {
     DiagnosticSeverity _severity;
     SourceLocation _location;
     std::string _message;
@@ -30,7 +30,7 @@ public:
     /// @param severity The severity level of the diagnostic.
     /// @param location The source location where the diagnostic occurred.
     /// @param message The message text.
-    DiagnosticMessage(
+    Diagnostic(
         DiagnosticSeverity severity, SourceLocation location,
         std::string message
     )
@@ -48,17 +48,17 @@ public:
     std::string const &getMessage() const { return _message; }
 };
 
-/// @class Diagnostic
+/// @class DiagnosticManager
 /// @brief Class for reporting and collecting diagnostics during compilation.
-class Diagnostic {
+class DiagnosticManager {
     SourceManager &_sourceManager;
-    llvm::SmallVector<DiagnosticMessage, 8> _messages;
+    llvm::SmallVector<Diagnostic, 8> _messages;
     bool _hasErrors = false;
 
 public:
-    /// @brief Constructs a Diagnostic instance.
+    /// @brief Constructs a DiagnosticManager instance.
     /// @param sourceManager The source manager to use for location information.
-    explicit Diagnostic(SourceManager &sourceManager)
+    explicit DiagnosticManager(SourceManager &sourceManager)
         : _sourceManager(sourceManager)
     {
     }
@@ -85,7 +85,7 @@ public:
 
     /// @brief Prints all collected diagnostics to the specified output stream.
     /// @param os The output stream where diagnostics will be printed.
-    void printAll(llvm::raw_ostream &os = llvm::errs()) const;
+    void printAll(llvm::raw_ostream &os = llvm::errs());
 
     /// @brief Returns whether any errors have been reported.
     /// @return True if any errors have been reported, false otherwise.
@@ -93,7 +93,7 @@ public:
 
     /// @brief Returns all collected diagnostic messages.
     /// @return A vector of all diagnostic messages.
-    llvm::SmallVector<DiagnosticMessage, 8> const &getMessages() const
+    llvm::SmallVector<Diagnostic, 8> const &getMessages() const
     {
         return _messages;
     }
@@ -110,8 +110,7 @@ private:
     /// @brief Formats and prints a single diagnostic message.
     /// @param os The output stream where the diagnostic will be printed.
     /// @param msg The diagnostic message to print.
-    void
-    printDiagnostic(llvm::raw_ostream &os, DiagnosticMessage const &msg) const;
+    void printDiagnostic(llvm::raw_ostream &os, Diagnostic const &msg) const;
 };
 
 } // namespace glu
--- a/include/Basic/Diagnostic.hpp
+++ b/include/Basic/Diagnostic.hpp
@@ -1,0 +1,119 @@
+#ifndef GLU_BASIC_DIAGNOSTIC_HPP
+#define GLU_BASIC_DIAGNOSTIC_HPP
+
+#include "SourceLocation.hpp"
+#include "SourceManager.hpp"
+#include <llvm/ADT/SmallVector.h>
+#include <llvm/ADT/StringRef.h>
+#include <llvm/Support/raw_ostream.h>
+#include <string>
+#include <vector>
+
+namespace glu {
+
+/// @brief Level of severity for diagnostic messages.
+enum class DiagnosticSeverity {
+    Note, ///< Informational message
+    Warning, ///< Warning that doesn't prevent compilation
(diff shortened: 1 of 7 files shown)


lib/Basic/Diagnostic.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+void Diagnostic::note(SourceLocation loc, llvm::StringRef message)
+{
+    addDiagnostic(DiagnosticSeverity::Note, loc, message);
+}
+
+void Diagnostic::fatal(SourceLocation loc, llvm::StringRef message)
+{
+    addDiagnostic(DiagnosticSeverity::Fatal, loc, message);
+    llvm::errs() << "Fatal error: compilation cannot continue.\n";
```
Snowy1803, line 36: Don't add a random extra print

lib/Basic/Diagnostic.cpp
```diff
@@ -0,0 +1,119 @@
...
+    size_t warningCount = 0;
+    size_t noteCount = 0;
+
+    for (auto const &msg : _messages) {
+        switch (msg.getSeverity()) {
+        case DiagnosticSeverity::Note: ++noteCount; break;
+        case DiagnosticSeverity::Warning: ++warningCount; break;
+        case DiagnosticSeverity::Error:
+        case DiagnosticSeverity::Fatal: ++errorCount; break;
+        default: break;
```
Snowy1803, line 99: Useless default would hide the warning if we add a new severity (-Wswitch)

lib/Basic/Diagnostic.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+        // Print a caret (^) pointing to the specific column
+        os.indent(column - 1);
+        os << "^\n";
+    }
+}
+
+void Diagnostic::printAll(llvm::raw_ostream &os) const
+{
+    for (auto const &msg : _messages) {
```
Snowy1803, line 84: Maybe the diagnostics should be sorted by source location

include/Basic/SourceManager.hpp
```diff
@@ -166,6 +166,21 @@ class SourceManager {
...
 
+    /// @brief Get the content of the buffer for a specific location
+    /// @param loc The source location
+    /// @return The content of the buffer as a StringRef
+    llvm::StringRef getBufferData(SourceLocation loc) const;
+
+    /// @brief Get the offset of the start of the line containing the location
+    /// @param loc The source location
+    /// @return The offset of the start of the line
+    size_t getLineStart(SourceLocation loc) const;
```
Snowy1803, line 177: If the return value is supposed to be a source location, the type should be SourceLocation

lib/Basic/SourceManager.cpp
```diff
@@ -225,3 +225,73 @@ void glu::SourceManager::reset()
     _nextOffset = 0;
     _mainFile = FileID(0);
 }
+
+llvm::StringRef glu::SourceManager::getBufferData(SourceLocation loc) const
```
Snowy1803, line 229: Difference with getCharacterData?

LindonAliu: nothing

lib/Basic/Diagnostic.cpp
LindonAliu: they're not added file by file (so sorted by default) ?

Snowy1803: They can be generated by different passes (the parser may have some diagnostics, then the global resolver, then the constraints, then GIL passes, etc), so no

lib/Basic/SourceManager.cpp
```diff
@@ -140,6 +140,13 @@ char const *glu::SourceManager::getCharacterData(SourceLocation loc) const
     return buffer.data() + offsetInFile;
 }
 
+llvm::StringRef
+glu::SourceManager::getCharacterDataInStringRef(SourceLocation loc) const
+{
+    auto *data = getCharacterData(loc);
+    return data ? llvm::StringRef(data) : llvm::StringRef();
```
Snowy1803, line 147: No, this will create a string ref from a null terminated big buffer which is extremely inefficient and completely useless. Keep the pointer.

include/Basic/Diagnostic.hpp
```diff
@@ -0,0 +1,119 @@
...
+/// @brief Level of severity for diagnostic messages.
+enum class DiagnosticSeverity {
+    Note, ///< Informational message
+    Warning, ///< Warning that doesn't prevent compilation
+    Error, ///< Error that prevents compilation
+    Fatal ///< Fatal error that stops compilation immediately
+};
+
+/// @brief A diagnostic message with location and severity information.
+class DiagnosticMessage {
```
Snowy1803, line 23: This should probably be called Diagnostic

include/Basic/Diagnostic.hpp
```diff
@@ -0,0 +1,119 @@
...
+    /// @return The source location where the diagnostic occurred.
+    SourceLocation getLocation() const { return _location; }
+
+    /// @return The message text.
+    std::string const &getMessage() const { return _message; }
+};
+
+/// @class Diagnostic
+/// @brief Class for reporting and collecting diagnostics during compilation.
+class Diagnostic {
```
Snowy1803, line 53: This should probably be called Diagnostics or DiagnosticManager or something

lib/Basic/Diagnostic.cpp
```diff
@@ -0,0 +1,137 @@
...
+namespace glu {
+
+void DiagnosticManager::addDiagnostic(
+    DiagnosticSeverity severity, SourceLocation loc, llvm::StringRef message
+)
+{
+    _messages.emplace_back(severity, loc, message.str());
+
+    // Set the error flag if this is an error or fatal error
+    _hasErrors = severity >= DiagnosticSeverity::Error;
```
Snowy1803, line 12: If hasErrors is true and we add a warning, this would set hasErrors back to false

lib/Basic/Diagnostic.cpp
```diff
@@ -0,0 +1,137 @@
...
+    if (_sourceManager.getCharacterData(msg.getLocation()) == nullptr) {
+        return;
+    }
+
+    auto lineStart = _sourceManager.getLineStart(msg.getLocation()).getOffset();
+    auto lineEnd = _sourceManager.getLineEnd(msg.getLocation()).getOffset();
+    auto column = _sourceManager.getSpellingColumnNumber(msg.getLocation());
+
+    // Print the line of source code
+    char const *data = _sourceManager.getCharacterData(msg.getLocation());
```
Snowy1803, line 72: This returns the data at the given location, not the start of the file

lib/Basic/Diagnostic.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+    // Print the actual message
+    os << msg.getMessage() << "\n";
+
+    if (_sourceManager.getCharacterData(msg.getLocation()) == nullptr) {
+        return;
+    }
+
+    auto lineStart = _sourceManager.getLineStart(msg.getLocation()).getOffset();
+    auto lineEnd = _sourceManager.getLineEnd(msg.getLocation()).getOffset();
```
Snowy1803, line 68: The getOffset function within source location should not exist, source locations are opaque data structures that can only be used by the source manager. The offset isn’t a real offset, there is no buffer in which you can use the offset to get a string out of it.

test/Basic/DiagnosticTest.cpp
```diff
@@ -0,0 +1,161 @@
...
+    size_t pos3 = output.find("Error line 4");
+
+    ASSERT_NE(pos1, std::string::npos);
+    ASSERT_NE(pos2, std::string::npos);
+    ASSERT_NE(pos3, std::string::npos);
+
+    // Check if the order is correct
+    ASSERT_LT(pos1, pos2);
+    ASSERT_LT(pos2, pos3);
+}
```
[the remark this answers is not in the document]
LindonAliu: It's a quick test to see if it works overall, I'm waiting to implement it in Parser to do functional tests

Snowy1803 (approved):

PR #350, opened 2025-05-13
 Feature: Create ConstraintSystem

### LINKED ISSUE #346: Create ConstraintSystem to store constraint, type var list, best solution, etc.

badermarouan:
Fix #346


include/Sema/ConstraintSystem.hpp
```diff
@@ -0,0 +1,70 @@
...
+#include <llvm/ADT/ilist.h>
+
+namespace glu::sema {
+
+class ConstraintSystem {
+    ScopeTable *_scopeTable; // The scope table for the current context
+    std::vector<glu::types::TypeVariableTy *>
+        _typeVariables; // List of type variables
+    llvm::BumpPtrAllocator _allocator; // Allocator for memory management
+    llvm::ilist<Constraint> _constraints; // List of constraints
```
Snowy1803, line 17: Constraints aren’t an intrusive list? We really only have one constraint which is a conjunction of all the others. Or a list of constraints directly. But it’s not intrusive, there’s no prev or next in class Constraint

Snowy1803 (approved):
lgtm

jeremyelalouf (approved):

PR #358, opened 2025-06-04
Feature: Generate type constraints for ReturnStmt

### LINKED ISSUE #331: Constrain ReturnStmt
Value convertible to function return type

LindonAliu:
Closes #331

### DIFF
--- a/lib/Sema/CSWalker.cpp
+++ b/lib/Sema/CSWalker.cpp
@@ -98,6 +98,22 @@ public:
             _cs.addConstraint(constraint);
         }
     }
+
+    /// @brief Visits a return statement and generates type constraints.
+    void postVisitReturnStmt(glu::ast::ReturnStmt *node)
+    {
+        auto *returnType = node->getReturnExpr()->getType();
+        auto *expectedReturnType = _cs.getScopeTable()
+                                       ->getFunctionDecl()
+                                       ->getType()
+                                       ->getReturnType();
+
+        _cs.addConstraint(
+            Constraint::createConversion(
+                _cs.getAllocator(), returnType, expectedReturnType, node
+            )
+        );
+    }
 };
 
 class GlobalCSWalker : public glu::ast::ASTWalker<GlobalCSWalker, void> {


lib/Sema/CSWalker.cpp
```diff
@@ -98,6 +98,22 @@ class LocalCSWalker : public glu::ast::ASTWalker<LocalCSWalker, void> {
             _cs.addConstraint(constraint);
         }
     }
+
+    /// @brief Visits a return statement and generates type constraints.
+    void postVisitReturnStmt(glu::ast::ReturnStmt *node)
+    {
+        auto *returnType = node->getReturnExpr()->getType();
```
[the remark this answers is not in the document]
LindonAliu: We have VoidTy

Snowy1803 (dismissed):
You should check how void is handled, as copilot says

LindonAliu:
> You should check how void is handled, as copilot says

Functions which return void have `VoidTy` in `returnType` i think ? So you choose whether to create a constraint. For the moment a constraint is created (but it is useless)

lib/Sema/CSWalker.cpp
```diff
@@ -98,13 +103,47 @@ class LocalCSWalker : public glu::ast::ASTWalker<LocalCSWalker, void> {
...
 };
 
 class GlobalCSWalker : public glu::ast::ASTWalker<GlobalCSWalker, void> {
     std::vector<ScopeTable> _scopeTable;
+    glu::DiagnosticManager &_diagManager;
 
 public:
-    GlobalCSWalker() { }
+    GlobalCSWalker(glu::DiagnosticManager &diagManager)
+        : _diagManager(diagManager)
```
badermarouan, line 144: should be a pointer

LindonAliu: after research, there is no difference in performance between passing a pointer or a reference (it's the same in assembler, only the address is passed)
https://cplusplus.com/forum/general/131984/#:~:text=Using%20a%20pointer%20vs,just%20initialized%20like%20a%20pointer

badermarouan (dismissed):

Snowy1803 (approved):

PR #387, opened 2025-06-10
 [Sema] Feature/ add postVisitUnaryOpExpr method to handle unary operation constraints

jeremyelalouf:


lib/Sema/CSWalker.cpp
```diff
@@ -205,6 +205,67 @@ class LocalCSWalker : public glu::ast::ASTWalker<LocalCSWalker, void> {
             )
         );
     }
+
+    void postVisitUnaryOpExpr(glu::ast::UnaryOpExpr *node)
```
Snowy1803, line 209: this is the same as a call, `+a` -> `` `+`(a)``, so no need to duplicate code. Difference is that we don't have to check for variables here, just functions.

lib/Sema/CSWalker.cpp
```diff
@@ -205,6 +205,67 @@ class LocalCSWalker : public glu::ast::ASTWalker<LocalCSWalker, void> {
...
+    {
+        auto *operandTy = node->getOperand()->getType();
+        auto *resultTy = node->getType();
+
+        auto &arena = node->getModule()->getContext()->getTypesMemoryArena();
+
+        auto *expectedFunctionTy = arena.create<glu::types::FunctionTy>(
+            llvm::ArrayRef<glu::types::TypeBase *>(&operandTy, 1), resultTy
```
Snowy1803, line 217:
```suggestion
            { operandTy }, resultTy
```
just use an initializer list...

lib/Sema/CSWalker.cpp
```diff
@@ -205,6 +205,67 @@ class LocalCSWalker : public glu::ast::ASTWalker<LocalCSWalker, void> {
...
+        );
+
+        llvm::StringRef opName = node->getOperator().getLexeme();
+        auto *item = _cs.getScopeTable()->lookupItem(opName);
+        llvm::ArrayRef<ast::DeclBase *> overloadDecls = item
+            ? llvm::makeArrayRef(item->decls)
```
Snowy1803, line 223:
```suggestion
            ? item->decls
```
just use the implicit conversion

lib/Sema/CSWalker.cpp
```diff
@@ -205,6 +205,67 @@ class LocalCSWalker : public glu::ast::ASTWalker<LocalCSWalker, void> {
...
+            : llvm::ArrayRef<ast::DeclBase *>();
```
Snowy1803, line 224:
```suggestion
            : {};
```

lib/Sema/CSWalker.cpp
```diff
@@ -205,6 +205,67 @@ class LocalCSWalker : public glu::ast::ASTWalker<LocalCSWalker, void> {
...
+
+        llvm::SmallVector<Constraint *, 4> constraints;
+
+        for (auto *decl : overloadDecls) {
+            if (auto *fnDecl = llvm::dyn_cast<ast::FunctionDecl>(decl)) {
+
+                if (fnDecl->getParamCount() != 1)
+                    continue;
```
Snowy1803, line 232: that works because we only consider function declarations with exactly one argument. but this doesn't use the constraint system, which means that the constraint resolution system doesn't know about the non-matches, and error messages will not have the whole picture.
To get better error messages, leave out those conditions, and let the solver handle it.

lib/Sema/CSWalker.cpp
```diff
@@ -205,6 +205,67 @@ class LocalCSWalker : public glu::ast::ASTWalker<LocalCSWalker, void> {
...
+
+                auto *fnTy = fnDecl->getType();
+
+                auto *constraintFunctionTy
+                    = arena.create<glu::types::FunctionTy>(
+                        fnTy->getParameters(), fnTy->getReturnType()
+                    );
```
Snowy1803, line 239: what? why don't you use `fnTy` directly??

lib/Sema/CSWalker.cpp
```diff
@@ -205,6 +205,67 @@ class LocalCSWalker : public glu::ast::ASTWalker<LocalCSWalker, void> {
...
+                        _cs.getAllocator(), expectedFunctionTy,
+                        constraintFunctionTy, node
+                    )
+                );
+            }
+        }
+
+        auto *builtinFunctionTy = arena.create<glu::types::FunctionTy>(
+            llvm::ArrayRef<glu::types::Ty> { operandTy }, operandTy
+        );
```
Snowy1803, line 252: what is that? that is not a thing?

lib/Sema/CSWalker.cpp
```diff
@@ -205,6 +205,67 @@ class LocalCSWalker : public glu::ast::ASTWalker<LocalCSWalker, void> {
...
+
+        constraints.push_back(
+            Constraint::createConversion(
+                _cs.getAllocator(), expectedFunctionTy, builtinFunctionTy, node
+            )
+        );
+
+        if (constraints.size() > 1) {
```
Snowy1803, line 260: what? createDisjunction already handles this case, don't add useless complexity. Do handle the `empty` case with a semantic error however.

lib/Sema/CSWalker.cpp
Snowy1803: the problem is that this is seriously hideous, you create a one-dimensional array from a variable on the stack
```suggestion
            std::initializer_list { operandTy }, resultTy
```
maybe? or
```suggestion
            llvm::ArrayRef<glu::types::TypeBase *> { operandTy }, resultTy
```

lib/Sema/CSWalker.cpp
```diff
@@ -205,6 +205,51 @@ class LocalCSWalker : public glu::ast::ASTWalker<LocalCSWalker, void> {
...
+                    )
+                );
+            }
+        }
+
+        if (constraints.empty()) {
+            _diagManager.error(
+                node->getLocation(),
+                "unary operator '" + node->getOperator().getLexeme().str()
+                    + "' has no candidate overloads"
```
Snowy1803, line 244: "use of undeclared operator 'XXX'"

lib/Basic/Diagnostic.cpp
```diff
@@ -18,24 +18,24 @@ void DiagnosticManager::addDiagnostic(
 #endif
 }
 
-void DiagnosticManager::error(SourceLocation loc, llvm::StringRef message)
+void DiagnosticManager::error(SourceLocation loc, llvm::Twine &&message)
```
Snowy1803, line 21: `llvm::Twine const &`

Snowy1803 (approved):

PR #388, opened 2025-06-30
Feature: Main loop to solve a ConstraintSystem

### LINKED ISSUE #382: Create Main resolver loop
Create a loop that iterates on the Constraints registered in the ConstraintSystem to solve them

badermarouan:
Fix #382

### DIFF
--- a/include/Sema/ConstraintSystem.hpp
+++ b/include/Sema/ConstraintSystem.hpp
@@ -134,10 +134,7 @@ struct SolutionResult {
     /// @brief Checks whether the result is ambiguous (i.e., multiple valid
     /// solutions exist).
     /// @return returns true if there are multiple solutions.
-    bool isAmbigous() const
-    {
-        return solutions.size() > 1;
-    }
+    bool isAmbigous() const { return solutions.size() > 1; }
 
     /// @brief Tries to add a new solution, checking for ambiguity and scoring.
     /// @param state The system state to convert and add as a solution.


include/Sema/ConstraintSystem.hpp
```diff
@@ -9,70 +9,332 @@
...
+    /// @brief Adds a new solution to the result.
+    /// @param s The solution to add (moved).
+    void addSolution(Solution &&s) { solutions.push_back(std::move(s)); }
+
+    /// @brief Checks whether any solutions were found.
+    /// @return True if at least one solution exists, false otherwise.
+    bool hasSolutions() const { return !solutions.empty(); }
+
+    /// @brief Marks the result as ambiguous.
+    void markAmbiguous() { isAmbiguous = true; }
```
Snowy1803, line 159: Why isn’t isAmbiguous just a function for solutions.size > 1?

badermarouan: Because there can be multiple solutions with different score that are valid. In this case we pick the lowest score solution. But it is ambiguous when we have multiple solutions with the same score.

include/Sema/ConstraintSystem.hpp
```diff
@@ -9,70 +9,240 @@
...
+
+    /// @brief The best solution found (the one with the lowest score).
+    Score bestScore = 0;
+
+    /// @brief Checks whether any solutions were found.
+    /// @return True if at least one solution exists, false otherwise.
+    bool hasSolutions() const { return !solutions.empty(); }
+
+    /// @brief Gets the best solution found (the one with the lowest score).
+    /// @return returns true if there are mutliple solutions.
```
Snowy1803, line 135: doc comment needs fixing

Snowy1803 (approved):

PR #397, opened 2025-07-08
Feature: Create TypeMapper

### LINKED ISSUE #336: Create Type Mapper (ASTVisitor)
This is an ASTVisitor that goes through all ASTNodes that contain types (see NODE_TYPEREF within NodeKind.def, note that it isn't up to date, and all Expressions now also have a type)

Its constructor takes two types. The visitor replaces all references of type A with type B.

Example:
```c++
TypeVariableTy *a;
IntTy *b;
TypeMapper(a, b)->visit(node); // TypeVariableTy is replaced with the known type
```

badermarouan:
Fix #336


include/AST/Types/TypeBase.hpp
```diff
@@ -13,6 +14,42 @@ enum class TypeKind {
...
+}
+
+/// @brief Enables string concatenation with a TypeKind in reverse order.
+/// @param kind A TypeKind to convert and prepend.
+/// @param rhs A std::string to append to.
+/// @return The resulting concatenated string.
+inline std::string operator+(glu::types::TypeKind kind, std::string const &rhs)
+{
+    return toString(kind) + rhs;
+}
```
Snowy1803, line 51: what is this for? wasn't this already somewhere?

badermarouan: It is used to generate diagnostic error messages. A similar function is used in the ASTPrinter but it overloads the operator '<<' instead of '+'

include/Sema/TypeMapper.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+#define NODE_TYPEREF(Type, Name)                                \
+    node->set##Name(this->asImpl()._mapType(node->get##Name()))
+
+#define NODE_CHILD(Type, Name) (void) 0
+#define NODE_CHILDREN(Type, Name) (void) 0
+
+    glu::types::TypeBase *_mapType(glu::types::TypeBase *type)
+    {
+        return mapType(type);
```
Snowy1803, line 22: asImpl

lib/Sema/TypeVariableTyMapper.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+    glu::types::TypeBase *visitTypeVariableTy(glu::types::TypeVariableTy *type)
+    {
+        auto mappedType = _solution->getTypeFor(type);
+        if (!mappedType) {
+            _diagManager.error(
+                SourceLocation::invalid,
+                // TODO: Use a TypeVisitor to print properly the type
+                std::string("Type variable mapping not found for: ")
+                    + type->getKind()
```
Snowy1803, line 36: this isn't even useful, the Kind here is by definition TypeVariableTy

include/Sema/TypeMapper.hpp
```diff
@@ -0,0 +1,33 @@
+#include "ASTWalker.hpp"
+#include "Types.hpp"
+
+namespace glu::sema {
+
+template <typename Impl>
+class TypeMapper : public glu::ast::ASTWalker<TypeMapper<Impl>, void> {
+    Impl *asImpl() { return static_cast<Impl &>(*this); }
```
Snowy1803, line 8: asImpl is already defined in ASTWalker, but you need to give it the same Impl
```suggestion
class TypeMapper : public glu::ast::ASTWalker<Impl, void> {
```

Snowy1803 (approved):

Snowy1803 (approved):

PR #408, opened 2025-07-10
Feature: apply bind

yunsecode:
Closes: #361

### DIFF
--- a/lib/Sema/ConstraintSystem.cpp
+++ b/lib/Sema/ConstraintSystem.cpp
@@ -80,13 +80,13 @@ bool ConstraintSystem::applyBind(Constraint *constraint, SystemState &state)
     if (llvm::isa<glu::types::TypeVariableTy>(first)
         && !llvm::isa<glu::types::TypeVariableTy>(second)) {
         state.typeBindings[static_cast<glu::types::TypeVariableTy *>(first)]
-            = nullptr;
+            = reinterpret_cast<glu::gil::Type *>(second);
         return true;
     }
     if (!llvm::isa<glu::types::TypeVariableTy>(first)
         && llvm::isa<glu::types::TypeVariableTy>(second)) {
         state.typeBindings[static_cast<glu::types::TypeVariableTy *>(second)]
-            = nullptr;
+            = reinterpret_cast<glu::gil::Type *>(first);
         return true;
     }
     if (llvm::isa<glu::types::TypeVariableTy>(first)
--- a/lib/Sema/ConstraintSystem.cpp
+++ b/lib/Sema/ConstraintSystem.cpp
@@ -77,22 +77,23 @@ bool ConstraintSystem::applyBind(Constraint *constraint, SystemState &state)
     if (first == second)
         return true;
 
-    if (glu::types::TypeVariableTy::classof(first)
-        && !glu::types::TypeVariableTy::classof(second)) {
+    if (llvm::isa<glu::types::TypeVariableTy>(first)
+        && !llvm::isa<glu::types::TypeVariableTy>(second)) {
         state.typeBindings[static_cast<glu::types::TypeVariableTy *>(first)]
-            = nullptr;
-        state.score += 1;
+            = static_cast<glu::gil::Type *>(second
+            ); // 타입 변수 바인딩을 second로
         return true;
     }
-    if (!glu::types::TypeVariableTy::classof(first)
-        && glu::types::TypeVariableTy::classof(second)) {
+    if (!llvm::isa<glu::types::TypeVariableTy>(first)
+        && llvm::isa<glu::types::TypeVariableTy>(second)) {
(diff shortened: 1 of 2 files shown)


lib/Sema/ConstraintSystem.cpp
```diff
@@ -63,16 +63,46 @@ void ConstraintSystem::solveConstraints()
...
 
+bool ConstraintSystem::applyBind(Constraint *constraint, SystemState &state)
+{
+    auto *first = constraint->getFirstType();
+    auto *second = constraint->getSecondType();
+
+    if (first == second)
+        return true;
+
+    if (glu::types::TypeVariableTy::classof(first)
```
Snowy1803, line 74: use llvm::isa not classof

lib/Sema/ConstraintSystem.cpp
```diff
@@ -63,16 +63,46 @@ void ConstraintSystem::solveConstraints()
...
+        && !glu::types::TypeVariableTy::classof(second)) {
+        state.typeBindings[static_cast<glu::types::TypeVariableTy *>(first)]
+            = nullptr;
```
Snowy1803, line 77: not nullptr, set it to `second`

lib/Sema/ConstraintSystem.cpp
```diff
@@ -63,16 +63,46 @@ void ConstraintSystem::solveConstraints()
...
+        state.score += 1;
```
Snowy1803, line 78: don't increase the score, only conversions increase the score

lib/Sema/ConstraintSystem.cpp
```diff
@@ -63,16 +63,44 @@ void ConstraintSystem::solveConstraints()
...
+    auto *first = constraint->getFirstType();
+    auto *second = constraint->getSecondType();
+
+    if (first == second)
+        return true;
+
+    if (llvm::isa<glu::types::TypeVariableTy>(first)
+        && !llvm::isa<glu::types::TypeVariableTy>(second)) {
+        state.typeBindings[static_cast<glu::types::TypeVariableTy *>(first)]
+            = reinterpret_cast<glu::gil::Type *>(second);
```
Snowy1803, line 77: you can't reinterpret_cast, the types are not compatible. Marouan's PR changes so that they are types::TypeBase and not gil::Type, so you can rebase on his PR to fix that.

lib/Sema/ConstraintSystem.cpp
```diff
@@ -63,16 +63,44 @@ void ConstraintSystem::solveConstraints()
...
+{
+    auto *first = constraint->getFirstType();
+    auto *second = constraint->getSecondType();
+
+    if (first == second)
+        return true;
+
+    if (llvm::isa<glu::types::TypeVariableTy>(first)
+        && !llvm::isa<glu::types::TypeVariableTy>(second)) {
+        state.typeBindings[static_cast<glu::types::TypeVariableTy *>(first)]
```
Snowy1803, line 76: don't use static_cast, use llvm::cast, or better, use llvm::dyn_cast in the if instead of llvm::isa to bind a variable

lib/Sema/ConstraintSystem.cpp
```diff
@@ -69,16 +69,42 @@ void ConstraintSystem::solveConstraints()
...
+bool ConstraintSystem::applyBind(Constraint *constraint, SystemState &state)
+{
+    auto *first = constraint->getFirstType();
+    auto *second = constraint->getSecondType();
+
+    if (first == second)
+        return true;
+
+    if (auto *firstVar = llvm::dyn_cast<glu::types::TypeVariableTy>(first);
+        firstVar && !llvm::dyn_cast<glu::types::TypeVariableTy>(second)) {
```
Snowy1803, line 81: Use isa when it’s a Boolean

lib/Sema/ConstraintSystem.cpp
```diff
@@ -69,16 +69,42 @@ void ConstraintSystem::solveConstraints()
...
+        return true;
+    }
+    if (auto *secondVar = llvm::dyn_cast<glu::types::TypeVariableTy>(second);
+        !llvm::dyn_cast<glu::types::TypeVariableTy>(first) && secondVar) {
+        state.typeBindings[secondVar] = first;
+        return true;
+    }
+    if (llvm::dyn_cast<glu::types::TypeVariableTy>(first)
+        && llvm::dyn_cast<glu::types::TypeVariableTy>(second)) {
+        return true;
```
Snowy1803, line 92: There is probably something to simplify here — maybe still bind one type to the other?

Snowy1803: we'll have to remember to handle this case in the solver simplification phase

Snowy1803 (approved):

Snowy1803:
Wait for the base branch to be merged before merging.

PR #424, opened 2025-07-11
Feature: Resolve Defaultable

yunsecode:
Closes: #364

### DIFF
--- a/lib/Sema/ConstraintSystem.cpp
+++ b/lib/Sema/ConstraintSystem.cpp
@@ -111,11 +111,6 @@ bool ConstraintSystem::applyDefaultable(
             appliedState.typeBindings[firstVar] = second;
             worklist.push_back(appliedState);
             return true;
-        } else if (existingBinding->second == second) {
-            return true;
-        } else {
-            worklist.push_back(state);
-            return true;
         }
     }
     return true;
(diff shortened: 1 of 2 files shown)


lib/Sema/ConstraintSystem.cpp
```diff
@@ -96,15 +96,43 @@ bool ConstraintSystem::applyBind(Constraint *constraint, SystemState &state)
...
+            worklist.push_back(appliedState);
+            return true;
+        } else if (existingBinding->second == second) {
+            return true;
+        } else {
+            worklist.push_back(state);
+            return true;
+        }
+    }
+    worklist.push_back(state);
```
Snowy1803, line 121: don't push to wordlist if you aren't changing anything

lib/Sema/ConstraintSystem.cpp
```diff
@@ -96,15 +96,42 @@ bool ConstraintSystem::applyBind(Constraint *constraint, SystemState &state)
...
+            SystemState appliedState = state;
+            appliedState.typeBindings[firstVar] = second;
+            worklist.push_back(appliedState);
+            return true;
+        } else if (existingBinding->second == second) {
+            return true;
+        } else {
+            worklist.push_back(state);
+            return true;
+        }
```
Snowy1803, line 119: remove both else if and else they're useless

PR #448, opened 2025-07-11
Sema Grand Unification Algorithm

### LINKED ISSUE #372: Resolve ValueMember / UnresolvedValueMember

### LINKED ISSUE #375: Resolve Disjunctions

### LINKED ISSUE #376: Resolve Conjunctions

Snowy1803:
Closes #372 (single test, could have more)
Closes #375 (single test, should have more)
Closes #376 (single test, could have more)


Snowy1803:
TODO:
 - Defaultable doesn't work: when one solution is missing a binding, it's still considered valid, which is WRONG
 - Conversions don't work: they need to use a recursive visitor that handles TypeVariableTy and work similarly to Unify

lib/Sema/Solver.cpp
```diff
@@ -46,9 +46,15 @@ bool SystemState::isFullyResolved(
     std::vector<Constraint *> const &constraints
 ) const
 {
-    for (Constraint *c : constraints)
-        if (!c->isDisabled() && c->isActive())
-            return false;
```
LindonAliu, line 51: i think its the good logic here

lib/Sema/ConstraintSystem.cpp
```diff
@@ -105,74 +113,528 @@ void ConstraintSystem::solveConstraints()
...
+    if (type == functionType) {
+        return ConstraintResult::Satisfied;
+    }
+
+    // Try to unify the type with the function type
+    if (unify(type, functionType, state)) {
+        // Record the overload choice in the state
+        SystemState newState = state;
+        // Note: We need the expression this applies to, which isn't available
+        // here This might need to be redesigned to pass the expression context
```
LindonAliu, line 305: thats true, maybe we need to create an issue for this

lib/Sema/Solver.cpp
Snowy1803: isActive is false by default so it was just failing always. also, we never touch those attributes of constraints, so it doesn't matter anyway

PR #451, opened 2025-07-12
Fix: correction of defaultable in sema

LindonAliu:


Snowy1803:
Cherry-picked into #448

Issue #489, opened 2025-09-02
Semantic Pass: Warn on Unused Variables


Snowy1803:
Add a semantic pass to generate a warning for declared but unused variables within functions or blocks. This helps maintain code cleanliness. Should ideally be a warning, not an error.

PR #493, opened 2025-09-02
Feature: Warn on Unused Variables

yunsecode:
Closes: #489


test/functional/Sema/unused_var.glu
```diff
@@ -0,0 +1,38 @@
...
+    flag;
+    return;
+}
+
+// CHECK-NOT: warning
+func noWarnUnusedVar() {
+    var flag: Bool = true;
+    flag;
+    return;
+}
```
Snowy1803, line 25: you could also check for assignments
```
var flag = true;
flag = false; // not an use
```

Snowy1803 (dismissed):

Issue #490, opened 2025-09-02
Semantic Pass: Detect Dead Code After Return/Break/Continue


Snowy1803:
Implement a semantic pass to detect unreachable code following a return, break, or continue statement. Should generate a warning when such code is found.

PR #496, opened 2025-09-02
Feature: Detect Dead Code

yunsecode:
Closes: #490


lib/Sema/CSWalker.cpp
```diff
@@ -4,6 +4,8 @@
 #include "AST/Types.hpp"
 #include "Sema/ConstraintSystem.hpp"
 
+#include "UnreachableWalker.cpp"
```
Snowy1803, line 7: Just rename the file to an hpp if you include it.

PR #550, opened 2025-09-13
[IRGen] Feature: Name Mangling

### LINKED ISSUE #533: /!\ P1: Add name mangling support and add attribute to disable it
Add name mangling so that functions with the same name but different namespace or type don't collide
Add ability to disable it for C function declarations (puts, etc) with an attribute (such as `@cdecl`)

Snowy1803:
Closes #533


lib/Sema/GlobalScopeVisitor.cpp
```diff
@@ -144,20 +144,20 @@ class GlobalScopeVisitor
...
+                = _scopeTable->getModule()->getImportName();
             bool isInDefaultImports = false;
-            for (llvm::sys::path::const_iterator it = llvm::sys::path::begin(moduleName),
-                                                end = llvm::sys::path::end(moduleName);
-                it != end; ++it) {
-                if (*it == "defaultImports") {
+            for (auto component : llvm::make_range(
+                     llvm::sys::path::begin(moduleName),
+                     llvm::sys::path::end(moduleName)
+                 )) {
```
LindonAliu, line 153: unreadable, move llvm:make_range in a variable before the loop

LindonAliu (commented):
lgtm

PR #569, opened 2025-09-23
Feature: ability to print nodes within LLDB

Snowy1803:
Debugging gluc is annoying without the ability to easily see what node we're looking at.
It is now easy to just do `p node->print()` in LLDB and see the output in the terminal

> [!NOTE]  
> Because these functions are only used in LLDB and not used in the project, they can't be inline in the .hpp or they will never be compiled and won't work
> Also, we can't use default arguments as those don't work in LLDB


LindonAliu (approved):
lgtm

PR #570, opened 2025-09-24
Feature: Check the validity of main

### LINKED ISSUE #549: Semantic Pass: Add checking for correct main function
Only allow correct main functions, similar to C:
```
func main() -> Void or Int;
func main(argc: Int, argv: **Char) -> Void or Int;
func main(argc: Int, argv: **Char, envp: **Char) -> Void or Int;
```

LindonAliu:
Closes #549


Snowy1803 (commented):
Good start. There should be a GIL pass to convert main return types from $Void to $Int, converting return instructions to return 0

PR #576, opened 2025-09-26
Feature: add checker to check duplicate function with no_mangling attribute

### LINKED ISSUE #572: Check for duplicates with @no_mangling
Create a semantic pass that checks whether we only have one toto function if it has a @no_mangling attribute.
Also add the @no_mangling attribute to the main function.

LindonAliu:
Closes #572

### DIFF
--- a/test/functional/Sema/multiple_functions_error.glu
+++ b/test/functional/Sema/multiple_functions_error.glu
@@ -7,9 +7,10 @@ func toto() {
     return;
 }
 
+// CHECK-NOT: error:
 func toto(a: Int, argv: **Char) -> Int {
     return a;
 }
 
-//CHECK: 15:1: error: duplicate function declaration: toto
+//CHECK: 16:1: error: duplicate function declaration: toto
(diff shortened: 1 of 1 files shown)


lib/Sema/SemanticPass/DuplicateFunctionNoManglingChecker.hpp
```diff
@@ -0,0 +1,63 @@
...
+    {
+        auto *attrs = node->getAttributes();
+
+        if (!attrs)
+            return false;
+        return llvm::any_of(attrs->getAttributes(), [](auto const *attr) {
+            return attr->getAttributeKind()
+                == ast::AttributeKind::NoManglingKind;
+        });
+    }
```
Snowy1803, line 58: it's just `node->getAttributes()->hasAttribute(ast::AttributeKind::NoManglingKind)` you don't need a function for that...

Snowy1803 (approved):

PR #579, opened 2025-09-26
Feature/ add ConstraintPrinter for printing constraints and enhance cons…

jeremyelalouf:
…trainAST with dump option


lib/Sema/ConstraintPrinter.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+    case ConstraintKind::ExpressibleByFloatLiteral:
+        result = formatType(constraint->getSingleType())
+            + " : ExpressibleByFloatLiteral";
+        break;
+
+    case ConstraintKind::ExpressibleByBoolLiteral:
+        result = formatType(constraint->getSingleType())
+            + " : ExpressibleByBoolLiteral";
+        break;
```
Snowy1803, line 209: not really a need to put the name of the constraint twice?

lib/Sema/ConstraintPrinter.cpp
```diff
@@ -0,0 +1,381 @@
...
+        result += "]";
+    }
+
+    // Add source location info
+    if (constraint->getLocator()) {
+        result += " @<AST:0x";
+        result += llvm::Twine::utohexstr(
+                      reinterpret_cast<uintptr_t>(constraint->getLocator())
+        )
+                      .str();
```
Snowy1803, line 270: can't you use ostream utils??

lib/Sema/ConstraintPrinter.cpp
```diff
@@ -0,0 +1,381 @@
...
+        = llvm::dyn_cast<glu::types::FunctionTy>(decl->getType())) {
+        result += "(";
+        auto params = funcType->getParameters();
+        for (size_t i = 0; i < params.size(); ++i) {
+            if (i > 0)
+                result += ", ";
+            result += formatType(params[i]);
+        }
+        result += ") -> " + formatType(funcType->getReturnType());
+    }
```
Snowy1803, line 306: all of that is just formatType(decl->getType())... stop remaking what already exists

lib/Sema/ConstraintPrinter.cpp
```diff
@@ -0,0 +1,363 @@
...
+            + std::to_string(constraint->getNestedConstraints().size())
+            + " conditions";
+        break;
+
+    case ConstraintKind::ExpressibleByIntLiteral:
+        result = formatType(constraint->getSingleType());
+        break;
+
+    case ConstraintKind::ExpressibleByStringLiteral:
+        result = formatType(constraint->getSingleType());
```
Snowy1803, line 195: duplicated code, use multiple case on the same statement

lib/Sema/ConstraintPrinter.cpp
```diff
@@ -0,0 +1,363 @@
...
+    Constraint const *constraint, llvm::raw_ostream &os, unsigned indent
+)
+{
+    if (!constraint) {
+        os.indent(indent);
+        os << "<null constraint>\n";
+        return;
+    }
+
+    constraint->print(os, indent);
```
Snowy1803, line 117: don't call this here, you lose `_typePrinter`
keep the implementation within this class

lib/Sema/ConstraintPrinter.cpp
```diff
@@ -0,0 +1,363 @@
...
+    /// @brief Recursively print a constraint and its nested constraints.
+    /// @param constraint The constraint to print.
+    /// @param os The output stream to print to.
+    /// @param indent The indentation level for formatting.
+    void printConstraintRecursive(
+        Constraint const *constraint, llvm::raw_ostream &os, unsigned indent
+    );
+};
+
+void Constraint::print(llvm::raw_ostream &os, unsigned indent) const
```
Snowy1803, line 66: this function should take no argument — it should only be used within LLDB

PR #597, opened 2025-10-02
Feature: Inline functions

### LINKED ISSUE #554: Make imports faster
- private imports within imported code (fastConstrainAST) can be ignored
- non-inline function content within imported code (fastConstrainAST) can be ignored

### LINKED ISSUE #588: Implement `@inline` attribute
This attribute on public function definitions allows other modules to also import the implementation of the function, so it can be optimized/inlined by LLVM.
It also changes the linkage of the function to use ODR (like the C++ inline keyword)

Snowy1803:
Closes #588: Implement inline attribute
Closes #554: Implement fastConstrainAST

Compiling a simple `test/functional/helloworld.glu` goes from 0.51s to 0.15s, a **70%** reduction in compile time. The reduction would be even greater for a file importing a lot of different modules.

This also adds support for delaying processing of private imports. This means that cyclic dependencies are now supported, as long as the imports are private.

### DIFF
--- a/include/AST/Attributes.def
+++ b/include/AST/Attributes.def
@@ -5,5 +5,6 @@
 
 ATTRIBUTE_KIND(NoMangling, "no_mangling", FunctionAttachment)
 ATTRIBUTE_KIND(CVariadic, "c_variadic", FunctionPrototypeAttachment)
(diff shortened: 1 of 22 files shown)


Snowy1803:
TODO: Using private imports within inline functions is broken

badermarouan (approved):

PR #598, opened 2025-10-02
Add attributes to all Declarations

### LINKED ISSUE #591: Add AttributeList to all DeclBase
AttributeList should be a member of DeclBase as all declarations can have attributes
AttributeChecker Sema pass should be updated to handle all kinds of declarations

Snowy1803:
Closes #591


LindonAliu (approved):
lgtm

PR #625, opened 2025-10-08
Feature: glu-demangle

### LINKED ISSUE #565: Create demangling tool
Create a new `glu-demangle` tool that takes a mangled glu function name and prints the demangled function signature
Different options could be added to customize how it's printed

LindonAliu:
Closes #565

### DIFF
--- a/tools/glu-demangle/src/args.rs
+++ b/tools/glu-demangle/src/args.rs
@@ -12,14 +12,14 @@ pub struct Args {
     #[arg(short, long)]
     pub strip_module: bool,
 
-    /// Read mangled names from stdin
+    /// Read mangled names from stdin [default if no names provided]
     #[arg(long)]
     pub stdin: bool,
 
     /// Mangled function names to demangle
     pub mangled_names: Vec<String>,
 
-    /// Output file (default: stdout)
+    /// Output file [default: stdout]
     #[arg(short, long)]
     pub output: Option<String>,
 }


edouard-sn:
hophophop

coverage.sh
```diff
@@ -35,6 +35,10 @@ LLVM_PROFILE_FILE="${PROFRAW_FILE}" ${UNIT_TESTS_EXEC} || error_exit "Unit tests
 
 echo -e "\nRunning functional tests with coverage instrumentation..."
 (cd ${FUNCTIONAL_TESTS_DIR} && LLVM_PROFILE_FILE="${PROFRAW_FILE}" ./${FUNCTIONAL_TESTS_EXEC}) || error_exit "Functional tests execution failed."
+
+echo -e "\nRunning glu-demangle tests..."
+(cd tools/glu-demangle && cargo test) || error_exit "glu-demangle tests execution failed."
```
Snowy1803, line 40: The tool itself is not built anywhere, add it to the cmake maybe? I'd want a binary in `build/tools/glu-demangle`

tools/glu-demangle/src/main.rs
```diff
@@ -0,0 +1,48 @@
...
+        Box::new(io::stdout())
+    };
+
+    if args.should_read_stdin() {
+        let stdin = io::stdin();
+        for line in stdin.lock().lines() {
+            let line = line?;
+            let trimmed = line.trim();
+            if !trimmed.is_empty() {
+                process_name(trimmed, &mut output, &args.format, args.strip_module)?;
```
Snowy1803, line 31: don't propagate the error out, let the user write new strings

tools/glu-demangle/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+///     assert_eq!(**pointee, GluType::Int { signed: true, width: 32 });
+/// }
+///
+/// // Invalid mangled name
+/// let result = demangle("INVALID$4main4testFvR");
+/// assert!(result.is_err());
+/// ```
+pub fn demangle(mangled: &str) -> anyhow::Result<GluFunction> {
+    if !mangled.starts_with("$GLU$") {
+        return Err(anyhow::anyhow!("Invalid GLU mangled name"));
```
Snowy1803, line 52: Maybe you could try parsing the rest of the string? When I copy paste from my terminal, it doesn't take the $GLU$ part, so you could maybe allow the user to not write it

tools/glu-demangle/src/formatter.rs
```diff
@@ -0,0 +1,230 @@
...
+///     element_type: Box::new(GluType::Int { signed: true, width: 32 })
+/// };
+/// assert_eq!(format_type(&array_type, false), "[10]i32");
+/// ```
+pub fn format_type(ty: &GluType, strip_module: bool) -> String {
+    match ty {
+        GluType::Void => "void".to_string(),
+        GluType::Bool => "bool".to_string(),
+        GluType::Char => "char".to_string(),
+        GluType::Null => "null".to_string(),
```
Snowy1803, line 146: Those types have names in PascalCase
```suggestion
        GluType::Void => "Void".to_string(),
        GluType::Bool => "Bool".to_string(),
        GluType::Char => "Char".to_string(),
        GluType::Null => "Null".to_string(),
```

tools/glu-demangle/src/formatter.rs
```diff
@@ -0,0 +1,230 @@
...
+        GluType::DynamicArray => "[]".to_string(),
+        GluType::Int { signed, width } => {
+            if *signed { format!("i{}", width) } else { format!("u{}", width) }
```
Snowy1803, line 149: You should use the user-facing names of those types
```suggestion
            if *signed { format!("Int{}", width) } else { format!("UInt{}", width) }
```

tools/glu-demangle/src/formatter.rs
```diff
@@ -0,0 +1,230 @@
...
+        }
+        GluType::Float { width } => format!("f{}", width),
```
Snowy1803, line 151:
```suggestion
        GluType::Float { width } => format!("Float{}", width),
```

tools/glu-demangle/src/formatter.rs
```diff
@@ -0,0 +1,230 @@
...
+        GluType::Pointer(pointee) => format!("*{}", format_type(pointee, strip_module)),
+        GluType::StaticArray { size, element_type } => {
+            format!("[{}]{}", size, format_type(element_type, strip_module))
```
Snowy1803, line 154: Wouldn't it be the opposite? But I don't think we have a syntax for them yet
```suggestion
            format!("{}[{}]", size, format_type(element_type, strip_module))
```

tools/glu-demangle/src/formatter.rs
```diff
@@ -0,0 +1,230 @@
...
+                .join(", ");
+            format!("({}) -> {}", params_str, format_type(return_type, strip_module))
+        }
+        GluType::Struct { module_path, name } | GluType::Enum { module_path, name } => {
+            if strip_module {
+                name.clone()
+            } else {
+                format!("{}::{}", module_path.join("::"), name)
+            }
+        }
```
Snowy1803, line 169: these look like format_type_only and format_name_only, why duplicate the code?

tools/glu-demangle/src/formatter.rs
```diff
@@ -0,0 +1,230 @@
...
+        func.name.clone()
+    } else {
+        format!("{}::{}", func.module_path.join("::"), func.name)
+    }
+}
+
+/// Format function type only
+fn format_type_only(func: &GluFunction) -> String {
+    let params = func.parameters.iter()
+        .map(|t| format_type(t, false))
```
Snowy1803, line 110: strip_module should be passed here:
```
% (cd ./tools/glu-demangle && cargo run -- -f=t --strip-module)
$GLU$3glu4file9writeFileFi32T3glu4file4FileSPci32R
(glu::file::File, *char, i32) -> i32
```
is wrong

tools/glu-demangle/src/formatter.rs
```diff
@@ -0,0 +1,230 @@
...
+    let params = func.parameters.iter()
+        .map(|t| format_type(t, strip_module))
+        .collect::<Vec<_>>()
+        .join(", ");
+
+    format!("{}({}) -> {}", name, params, format_type(&func.return_type, strip_module))
+}
+
+/// Format function in compact style
+fn format_compact(func: &GluFunction, strip_module: bool) -> String {
```
Snowy1803, line 50: compact is the same as signature with an extra `: `
looks useless?

PR #650, opened 2025-10-12
[Sema] Feature: CS Scoring

### LINKED ISSUE #609: Cannot assign negative integer literal to typed variable (Int) at top level
Assigning a negative integer literal to a typed variable like `Int` causes an ambiguous type resolution error when done outside a function context.

// OK
fn foo(x: Int) {
}

foo(-456);

// No OK
varr b: Int = -456;

### LINKED ISSUE #618: Fix Constraint System to use Scoring system
The constraint system currently does not use a scoring system, and instead relies on the correct order of constraints, which can be flaky

Snowy1803:
Closes #618
Works but currently slows Sema down by a lot
Closes #609 as well

### DIFF
--- a/include/Sema/Sema.hpp
+++ b/include/Sema/Sema.hpp
@@ -10,6 +10,11 @@ namespace glu::sema {
 class ImportManager;
 class ScopeTable;
 
+/// @brief Checks if a type is concrete (not a type variable or dependent on
+/// type variables).
+/// @param type The type to check.
+/// @return True if the type is concrete (no type variables left), false
+/// otherwise.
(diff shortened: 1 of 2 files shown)


LindonAliu (dismissed):
lgtm

include/Sema/Sema.hpp
```diff
@@ -10,6 +10,8 @@ namespace glu::sema {
 class ImportManager;
 class ScopeTable;
 
+bool typeIsConcrete(glu::types::Ty type);
```
badermarouan: doc, define what is a concrete type

PR #653, opened 2025-10-15
[Sema] Refactor: Remove expressions from solveConstraints

Snowy1803:
This was only used for tests, and those can just use `setRoot` instead.


LindonAliu (approved):
lgtm

Issue #623, opened 2025-10-06
Create DCE GIL Pass


Snowy1803:
Create a Dead Code Elimination pass.
Any unreachable basic block (blocks which are never branched to (`cond_br` or `br` towards it) can be removed.
If the removed instructions have a source location, we can generate a warning about the code being unreachable.

PR #655, opened 2025-10-15
Feature: add GilPass return last checker

### LINKED ISSUE #624: Replace ReturnLastChecker with a GIL Pass
After DCE, a GIL Pass should run that detects any `unreachable` GIL instruction that is in fact reachable (because it runs after DCE, all basic blocks are reachable, so no `unreachable` instruction should remain).
Diagnose an error for any `UnreachableInst` that remains. Remove ReturnLastChecker as this new system is more accurate (it will correctly determine if we return from all if-else paths, for instance)

```swift
func correct() -> Int {
  if cond {
    return 1;
  } else {
    return 0;
  }
  // no diagnostic even if there is no return at the end, because all paths return
}
```

LindonAliu:
Closes #624 #623

### DIFF
--- a/include/Basic/SourceLocation.hpp
+++ b/include/Basic/SourceLocation.hpp
@@ -108,6 +108,25 @@ template <> struct DenseMapInfo<glu::FileID> {
     }
     static bool isEqual(glu::FileID LHS, glu::FileID RHS) { return LHS == RHS; }
 };
+
+template <> struct DenseMapInfo<glu::SourceLocation> {
+    static glu::SourceLocation getEmptyKey()
+    {
+        return glu::SourceLocation(~0U);
+    }
+    static glu::SourceLocation getTombstoneKey()
+    {
+        return glu::SourceLocation(~0U - 1);
+    }
+    static unsigned getHashValue(glu::SourceLocation val)
+    {
+        return llvm::hash_value(val.getOffset());
+    }
+    static bool isEqual(glu::SourceLocation LHS, glu::SourceLocation RHS)
+    {
+        return LHS == RHS;
+    }
+};
 } // namespace llvm
 
 #endif // GLU_SOURCE_LOCATION_HPP


lib/GILGen/GILPasses/DeadCodeElimination.hpp
```diff
@@ -0,0 +1,145 @@
...
+                    successors.push_back(elseBlock);
+            }
+            // ReturnInst and UnreachableInst have no successors
+
+            for (auto *successor : successors) {
+                auto [_, inserted] = reachableBlocks.insert(successor);
+                if (inserted) {
+                    worklist.push_back(successor);
+                }
+            }
```
Snowy1803: no?

lib/GILGen/GILPasses/DeadCodeElimination.hpp
```diff
@@ -0,0 +1,145 @@
...
+/// to). Any unreachable basic block is removed from the function.
+/// If the removed instructions have a source location, we generate a warning
+/// about the code being unreachable.
+class DeadCodeEliminationPass
+    : public gil::InstVisitor<DeadCodeEliminationPass> {
+private:
+    DiagnosticManager &diagManager;
+    llvm::DenseSet<gil::BasicBlock *> reachableBlocks;
+    llvm::SmallVector<gil::BasicBlock *, 8> blocksToRemove;
+    llvm::SmallVector<SourceLocation, 8> warnedLocations;
```
Snowy1803: why is this not a DenseSet

tools/gluc/sources/CompilerDriver.cpp
```diff
@@ -296,7 +296,7 @@ int CompilerDriver::processPreCompilationOptions()
         return 0;
     }
 
-    gilgen.runGILPasses(*_gilModule, _GILFuncArena);
+    gilgen.runGILPasses(*_gilModule, _GILFuncArena, *_diagManager);
```
Snowy1803: after printgil before irgen: 
if (_diagManager->hasErrors()) {
        return 1;
    }

lib/GILGen/GILPasses/DeadCodeElimination.hpp
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    void beforeVisitFunction(gil::Function *func)
+    {
+        blocksToRemove.clear();
+        warnedLocations.clear();
+        computeReachableBlocks(func);
+
+        // Collect unreachable blocks
+        for (auto &bb : func->getBasicBlocks()) {
```
Snowy1803: this could be in beforeVisitBasicBlock

PR #663, opened 2025-10-16
Feature: add a PassManager

LindonAliu:


lib/Optimizer/PassManagerOptions.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+llvm::cl::list<std::string> _printBeforePasses(
+    "print-gil-before-pass", llvm::cl::desc("Print GIL before specific pass"),
+    llvm::cl::ZeroOrMore, llvm::cl::value_desc("pass-name")
+);
+
+llvm::cl::list<std::string> _printAfterPasses(
+    "print-gil-after-pass", llvm::cl::desc("Print GIL after specific pass"),
+    llvm::cl::ZeroOrMore, llvm::cl::value_desc("pass-name")
+);
```
Snowy1803: you could add a ``--print-after-all``

include/Optimizer/Optimizer.hpp
```diff
@@ -1,5 +1,43 @@
...
+namespace glu {
+
+class DiagnosticManager;
+class SourceManager;
+
+namespace gil {
+class Module;
+}
+
 class Optimizer {
```
Snowy1803: I'm not sure you need separate Optimizer and PassManager classes, you could merge this into the `PassManager` class

lib/Optimizer/PassManager.cpp
```diff
@@ -0,0 +1,50 @@
...
+    DiagnosticManager &diagManager, SourceManager &sourceManager,
+    llvm::raw_ostream &output, gil::Module *module,
+    llvm::BumpPtrAllocator &GILFunctionsArena
+)
+    : _diagManager(diagManager)
+    , _sourceManager(sourceManager)
+    , _output(output)
+    , _printer(&sourceManager, output)
+    , _module(module)
+    , _GILFunctionsArena(GILFunctionsArena)
```
Snowy1803: var names shouldn't start with capital, just `arena` and `_arena` is fine (or take a gilgen::GlobalContext?)

PR #676, opened 2025-10-23
Feature: Add @linkage_name attribute

### LINKED ISSUE #670: Add @linkage_name attribute
New attribute for functions (both prototypes and definitions)
`@linkage_name("hello")` will change the linkage name of the function without changing the import name
Example:
```swift
@linkage_name("hello") func world();

let a = world();
```
The call to the function is spelled world, but the linker will look for a function named hello.
It is a more powerful version of @no_mangling (no_mangling sets the linkage name to the function name), so they are incompatible:
`@linkage_name("hello") func hello();` is the same as `@no_mangling func hello();`

Needed for IRDec

LindonAliu:
Closes #670


lib/Sema/SemanticPass/ValidAttributeChecker.hpp
```diff
@@ -51,6 +51,44 @@ class ValidAttributeChecker
...
+
+        // Check if linkage name is non-empty
+        if (linkageName.empty()) {
+            _diagManager.error(
+                attr->getLocation(), "Linkage name cannot be empty"
+            );
+        }
+
+        // TODO: Add more validation for valid linkage names if needed
+        // (e.g., check for valid identifier characters)
```
Snowy1803: remove todo

test/functional/Sema/linkage_name_valid.glu
```diff
@@ -0,0 +1,46 @@
+//
```
Snowy1803: This should probably be within an `IRGen/` with a CHECK for the output LLVM IR

Snowy1803 (approved):

PR #679, opened 2025-10-24
Feature: add Basic CopyLoweringPass

### LINKED ISSUE #708: Basic CopyLoweringPass

badermarouan:
Fix #708

### DIFF
--- a/lib/Optimizer/GILPasses/CopyLoweringPass.cpp
+++ b/lib/Optimizer/GILPasses/CopyLoweringPass.cpp
@@ -69,6 +69,8 @@ public:
         if (nextInst && llvm::isa<gil::StoreInst>(nextInst)) {
             auto *storeInst = llvm::cast<gil::StoreInst>(nextInst);
             if (storeInst->getSource() == loadInst->getResult(0)) {
+                // TODO: Replace with inst->replaceAllUsesWith() when
+                // implemented
                 storeInst->setSource(callInst->getResult(0));
             }
         }


lib/Optimizer/GILPasses/CopyLoweringPass.cpp
```diff
@@ -0,0 +1,131 @@
...
+            return;
+
+        if (!ctx)
+            return;
+
+        // Generate code to call the copy function if it exists
+        auto *structure = llvm::dyn_cast<types::StructTy>(
+            loadInst->getResultType(0).getType()
+        );
+        if (structure && structure->getDecl()->hasOverloadedCopyFunction()) {
```
Snowy1803: that can also be a guard-if

lib/Optimizer/GILPasses/CopyLoweringPass.cpp
```diff
@@ -0,0 +1,131 @@
...
+            inst->eraseFromParent();
+        }
+    }
+
+    void visitLoadInst(gil::LoadInst *loadInst)
+    {
+        // Don't transform loads inside copy functions (would cause infinite
+        // recursion)
+        if (inCopyFunction)
+            return;
```
Snowy1803: you would need `copy` functions to use a "borrowed" reference so that ownership isn't handled. It might have to be done from a pointer (which means you would only have `load [copy]` and no copy instructions?).

Snowy1803 (dismissed):

PR #706, opened 2025-11-06
Feature: Parse Template Declarations

### LINKED ISSUE #592: Parse Function Template Definition
Support having generic functions and structures in the AST

LindonAliu:
Closes #592


lib/ASTPrinter/ASTPrinter.cpp
```diff
@@ -1,3 +1,4 @@
+#include "AST/Decl/TemplateParameterDecl.hpp"
```
Snowy1803: useless

PR #719, opened 2025-11-10
Fix: memory leaks

### LINKED ISSUE #717: Fix all memory leaks in compiler

LindonAliu:
Closes #717

### DIFF
--- a/include/GIL/Module.hpp
+++ b/include/GIL/Module.hpp
@@ -36,8 +36,6 @@ public:
     {
     }
 
-    ~Module();
-
     /// deleted copy constructor for a const Module instance
     Module(Module const &) = delete;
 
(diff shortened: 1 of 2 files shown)


lib/Sema/ModuleWalker.cpp
```diff
@@ -127,7 +129,8 @@ class ModuleWalker : public glu::ast::ASTWalker<ModuleWalker, void> {
     {
         if (_skippingCurrentFunction)
             return;
-        _scopeTable = new (_scopeTableAllocator) ScopeTable(_scopeTable, node);
+        _scopeTable = new (_scopeTableAllocator.Allocate())
+            ScopeTable(_scopeTable, node);
```
Snowy1803: Scope Tables for FunctionDecl and CompoundStmt are local — they probably shouldn't be allocated on the importmanager's scope table allocator because they are not used outside of ModuleWalker. Otherwise, those scope tables are kept for the whole lifetime of the compiler even when they are only used when applying sema on a single function. You could use a separate allocator within the ModuleWalker.

Snowy1803 (dismissed):

PR #734, opened 2025-11-13
Feature: implement ranges

### LINKED ISSUE #559: Implement range operators
Implement integer ranges: a..<b, creating a Range structure which has a begin iterator and an end iterator

instead of
```
for i in 0..<10 {
  std::print(i);
}
```
this
```
let range = 0..<10;
var it = std::begin(range);
while (it != std::end(range)) {
  let i = it.*;
  std::print(i);
  it = std::next(it);
}
```

`..<` is for 0 to 10, 10 exclus (0 to 9)
`...` is for 0 to 10, 10 inclus (0 to 10)

LindonAliu:
Closes #559

### DIFF
--- a/test/functional/run/range.glu
+++ b/test/functional/run/range.glu
@@ -12,7 +12,8 @@ func main() -> Int {
         iterator = std::next(iterator);
     }
 
-    std::printout("|");
+    // CHECK: 0123
+    std::print("");
 
     let exclusive = 2..<5;
     iterator = std::begin(exclusive);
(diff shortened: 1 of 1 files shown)


stdlib/defaultImports/range.glu
```diff
@@ -0,0 +1,65 @@
...
+}
+
+@inline public func ...(start: Int, end: Int) -> Range {
+    return makeRange(start, end, true);
+}
+
+public func begin(range: Range) -> RangeIterator {
+    if (isEmpty(range)) {
+        return makeIterator(endValue(range));
+    }
```
Snowy1803: useless, if it's empty, range.start == endValue anyway

Issue #439, opened 2025-07-11
Implement For Statement Code Generation in GILGenStmt


LindonAliu:
## Description

The `visitForStmt` method in `GILGenStmt.cpp` currently contains a TODO and is not implemented. Future developers should implement code generation for the `ForStmt` AST node, enabling support for `for` loops in the GIL IR generation phase.

### Requirements
- Implement the logic in `visitForStmt` to generate GIL code for `ForStmt` nodes.
- Ensure correct handling of loop initialization, condition, increment, and body.
- Integrate with scope management and break/continue destinations as done for `WhileStmt`.
- Add appropriate error handling and unreachable code generation if needed.
- Update or add tests to verify correct GIL output for `for` loops.

### References
- See the implementation of `visitWhileStmt` for guidance on loop code generation structure.
- The current code contains commented-out scaffolding that may be useful as a starting point.

---

**Location:** `lib/GILGen/GILGenStmt.cpp`, method `visitForStmt`
**Current Status:** Not implemented (asserts false)

PR #737, opened 2025-11-15
Feature: For Loops

LindonAliu:
Closes: #439

### DIFF
--- a/lib/Sema/ConstraintSystem/ConstraintSystem.cpp
+++ b/lib/Sema/ConstraintSystem/ConstraintSystem.cpp
@@ -13,7 +13,6 @@ ConstraintSystem::ConstraintSystem(
 )
     : _scopeTable(scopeTable)
     , _root(scopeTable->getNode())
-    , _typeVariables()
     , _allocator()
     , _diagManager(diagManager)
     , _context(context)
--- a/lib/Sema/ConstraintSystem/LocalCSWalker.cpp
+++ b/lib/Sema/ConstraintSystem/LocalCSWalker.cpp
@@ -48,7 +48,6 @@ public:
             auto typeVar = _astContext->getTypesMemoryArena()
                                .create<glu::types::TypeVariableTy>();
             node->setType(typeVar);
-            _cs.addTypeVariable(typeVar);
         }
     }
 
@@ -239,17 +238,10 @@ public:
         auto &typesArena = _astContext->getTypesMemoryArena();
 
         if (!binding->getType()) {
-            auto *typeVar = typesArena.create<glu::types::TypeVariableTy>();
-            _cs.addTypeVariable(typeVar);
-            binding->setType(typeVar);
-        } else if (auto *typeVar = llvm::dyn_cast<glu::types::TypeVariableTy>(
-                       binding->getType()
-                   )) {
-            _cs.addTypeVariable(typeVar);
+            binding->setType(typesArena.create<glu::types::TypeVariableTy>());
         }
 
         auto *iteratorType = typesArena.create<glu::types::TypeVariableTy>();
-        _cs.addTypeVariable(iteratorType);
 
         auto *rangeType = range->getType();
 
@@ -317,7 +309,6 @@ public:
             auto *typeVar = _astContext->getTypesMemoryArena()
                                 .create<glu::types::TypeVariableTy>();
             varLet->setType(typeVar);
-            _cs.addTypeVariable(typeVar);
             varType = typeVar;
         }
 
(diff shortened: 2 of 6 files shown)


lib/Sema/ConstraintSystem/ConstraintSystem.cpp
```diff
@@ -457,7 +458,9 @@ ConstraintSystem::applyBindOverload(Constraint *constraint, SystemState &state)
     auto *parent = constraint->getLocator()->getParent();
     bool needsFuncPtr = true;
 
-    if (auto *callExpr = llvm::dyn_cast<glu::ast::CallExpr>(parent)) {
+    if (!parent) {
```
Snowy1803: that shouldn't be possible

lib/Sema/ConstraintSystem/LocalCSWalker.cpp
```diff
@@ -228,9 +229,77 @@ class LocalCSWalker : public glu::ast::ASTWalker<LocalCSWalker, void> {
...
+            = [&](llvm::StringRef name) -> glu::ast::RefExpr * {
+            auto *ref
+                = _astContext->getASTMemoryArena().create<glu::ast::RefExpr>(
+                    node->getLocation(),
+                    glu::ast::NamespaceIdentifier {
+                        llvm::ArrayRef<llvm::StringRef>(), name },
+                    nullptr
+                );
+            preVisitExprBase(ref);
+            visit(ref);
```
Snowy1803: don't visit it twice

lib/Sema/ConstraintSystem/LocalCSWalker.cpp
```diff
@@ -228,9 +229,77 @@ class LocalCSWalker : public glu::ast::ASTWalker<LocalCSWalker, void> {
...
+        auto addHelperConstraint =
+            [&](glu::ast::RefExpr *ref,
+                llvm::ArrayRef<glu::types::TypeBase *> params,
+                glu::types::TypeBase *result) {
+                auto *fnTy
+                    = typesArena.create<glu::types::FunctionTy>(params, result);
+                _cs.addConstraint(
+                    Constraint::createConversion(_cs.getAllocator(), ref, fnTy)
+                );
+            };
```
Snowy1803: the function boilerplate is longer than the actual function...

lib/Sema/ConstraintSystem/LocalCSWalker.cpp
```diff
@@ -228,9 +229,77 @@ class LocalCSWalker : public glu::ast::ASTWalker<LocalCSWalker, void> {
...
+
+        auto *beginRef = createHelperRef("begin");
+        node->setBeginFunc(beginRef);
```
Snowy1803: you need to set it before visiting it, parent should never be nullptr within the CS

lib/Sema/ConstraintSystem/LocalCSWalker.cpp
```diff
@@ -424,7 +492,11 @@ class LocalCSWalker : public glu::ast::ASTWalker<LocalCSWalker, void> {
         ast::RefExpr *node, llvm::SmallVector<Constraint *, 4> &constraints
     )
     {
-        auto *parent = llvm::dyn_cast<ast::UnaryOpExpr>(node->getParent());
+        auto *parentNode = node->getParent();
```
Snowy1803: all of these shouldn't exist and are dangerous

lib/Sema/ConstraintSystem/LocalCSWalker.cpp
```diff
@@ -228,9 +229,77 @@ class LocalCSWalker : public glu::ast::ASTWalker<LocalCSWalker, void> {
         _cs.addConstraint(constraint);
     }
 
-    void _visitForStmt([[maybe_unused]] glu::ast::ForStmt *node)
+    void _visitForStmt(glu::ast::ForStmt *node)
     {
-        // TODO: #439 Implement for-loops
+        auto *binding = node->getBinding();
+        auto *range = node->getRange();
+
```
Snowy1803: you need to set the root of the CS in such a way that it doesn't include the body of the for loop, similar to the while loop (it needs to go over the rest of the for loop content, however
Or, you can override _visitForStmt in `TypeVariableTyMapper` so it doesn't go over the body
Or, fix the literal expressions to have a null type instead of type variables, just like other expressions.
Make sure this code compiles:
```
for i in 0..<10 {
 let j = i + 1;
}
```

Snowy1803 (approved):

PR #738, opened 2025-11-15
Feature: Local namespaces

### LINKED ISSUE #721: Implement namespace syntax
```
namespace thing { ... }
namespace flip::flop { ... }
```
Add keyword to Lexer, AST node and parse, and support in Sema

Cannot add to/replace an imported namespace, but can supplement a previous namespace.
```
import std;
namespace std {
  // nope. there is no right ScopeTable where to put this
}
```

```
namespace machin {}
namespace machin {} // ok. add to existing namespace
```

It could be used for adding fake cases to enums as well:
```
enum Variant { green, red, black }
namespace Variant {
    let default: Variant = green;
}

// Variant::green and Variant::default are now equivalent
```

Snowy1803:
Closes #721

### DIFF
--- a/include/GILGen/GILGen.hpp
+++ b/include/GILGen/GILGen.hpp
@@ -27,6 +27,11 @@ gil::Global *generateGlobal(
     gil::Module *module, ast::VarLetDecl *decl, GlobalContext &globalCtx
 );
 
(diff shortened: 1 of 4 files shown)


LindonAliu (dismissed):
lgtm

PR #762, opened 2025-11-29
Feature: Function pointer call

### LINKED ISSUE #672: Allow Function Pointer call in Sema
Allow pointer to functions to be called. It seems to already be implemented in GILGen & IRGen, so it should work out of the box (with #671 as well)

LindonAliu:
Closes #672

### DIFF
--- a/lib/Sema/ConstraintSystem/ConstraintSystem.cpp
+++ b/lib/Sema/ConstraintSystem/ConstraintSystem.cpp
@@ -37,15 +37,6 @@ static void insertImplicitCast(
     if (expr->getType() == targetType)
         return; // No conversion needed
 
-    // Don't insert a cast for function pointer to function type conversion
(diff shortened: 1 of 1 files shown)


include/GILGen/Context.hpp
```diff
@@ -321,14 +321,19 @@ class Context {
     gil::CallInst *
     buildCall(gil::Value functionPtr, llvm::ArrayRef<gil::Value> args)
     {
-        types::PointerTy pointerType
-            = llvm::cast<types::PointerTy>(functionPtr.getType().getType());
-        types::FunctionTy *funcType
-            = llvm::cast<types::FunctionTy>(pointerType.getPointee());
+        types::FunctionTy *funcType = nullptr;
```
Snowy1803: never nullptr initialize something that is definitely initialized, it hides warnings and can hide bugs

Snowy1803 (approved):

PR #788, opened 2025-12-04
Re-enable CopyLoweringPass

### LINKED ISSUE #787: Fix CopyLoweringPass
CopyLoweringPass required RAUW to be finished.

badermarouan:
Fix #787

### DIFF
--- a/lib/GILGen/GILGenStmt.cpp
+++ b/lib/GILGen/GILGenStmt.cpp
@@ -337,15 +337,10 @@ private:
 
     void dropScopeVariables(Scope &scope)
     {
-        for (auto &[var, val] : scope.getVariables()) {
-            if (llvm::isa<ast::ParamDecl>(var)
-                && ctx.getASTFunction()->getName() == "drop")
-                continue; // Don't drop parameters in drop functions, otherwise
-                          // infinite recursion
-            ctx.buildDropPtr(var->getType(), val);
-        }
-        for (auto &val : llvm::reverse(scope.getUnnamedAllocations())) {
-            ctx.buildDropPtr(val.getType(), val);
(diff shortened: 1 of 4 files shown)


lib/Optimizer/LatePasses/CopyLoweringPass.cpp
```diff
@@ -37,6 +40,24 @@ class CopyLoweringPass : public gil::InstVisitor<CopyLoweringPass> {
             return;
         }
 
+        // Check for infinite recursion: if we're inside the copy function
+        // for this struct, warn about potential infinite recursion
+        ast::FunctionDecl *copyFunc = structure->getDecl()->getCopyFunction();
+        if (_currentFunction && _currentFunction->getDecl() == copyFunc) {
```
Snowy1803: Can't you use loadInst->getParent()->getParent() instead of storing the _currentFunction as a field?

test/functional/Sema/copy_overload_infinite_recursion.glu
```diff
@@ -0,0 +1,35 @@
...
+// No warning here - using a helper function with pointer parameter
+func copy(r: *SafeResource) -> SafeResource {
+    return helperByPtr(r); // Pass by pointer, no implicit copy
+}
+
+func helperByPtr(r: *SafeResource) -> SafeResource {
+    return { r.*.data };
+}
+
+func main() {
```
Snowy1803: you don't need a main if you compile with `-c` just like all other Sema tests do

lib/Optimizer/LatePasses/CopyLoweringPass.cpp
Snowy1803: ?????????? no

Issue #782, opened 2025-12-03
Implement global destructors


Snowy1803:
Call global vars drop function at program end, when they are non-trivial and if they were initialized

PR #789, opened 2025-12-04
Feature: Global Destruction

LindonAliu:
Closes: #782


test/functional/run/global_destructor.glu
```diff
@@ -0,0 +1,29 @@
+// RUN: gluc %s -o %t && %t | FileCheck %s
```
Snowy1803: Add a test where the global is never initialized (shouldn't be destroyed), and a test for the above edge case

Snowy1803 (commented):
lgtm otherwise

Snowy1803 (approved):

PR #797, opened 2025-12-12
Fix: Visibility Checks

### LINKED ISSUE #681: Add missing checks for member visibility (struct field, functions from imported namespaces, enum cases, etc.)

badermarouan:
Fix #681

### DIFF
--- a/lib/Sema/ConstraintSystem/ConstraintSystem.cpp
+++ b/lib/Sema/ConstraintSystem/ConstraintSystem.cpp
@@ -564,6 +564,18 @@ ConstraintSystem::applyValueMember(Constraint *constraint, SystemState &state)
     auto const &field = structType->getField(*fieldIndex);
     auto *fieldType = field->getType();
 
+    // Check visibility: private fields cannot be accessed from other modules
+    auto *currentModule = _scopeTable->getModule();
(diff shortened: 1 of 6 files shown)


lib/Sema/ConstraintSystem/ConstraintSystem.cpp
```diff
@@ -564,6 +564,18 @@ ConstraintSystem::applyValueMember(Constraint *constraint, SystemState &state)
...
+    // Check visibility: private fields cannot be accessed from other modules
+    auto *currentModule = _scopeTable->getModule();
+    auto *structModule = structType->getDecl()->getModule();
+    if (field->isPrivate() && structModule && structModule != currentModule) {
+        _diagManager.error(
+            memberExpr->getLocation(),
+            llvm::Twine("Cannot access private field '") + memberName
+                + "' of struct '" + structType->getName() + "'"
+        );
+        return ConstraintResult::Failed;
```
Snowy1803: You are within the constraint system, so this function will be called multiple times if there are multiple overloads within the statement.
So either:
- Put the diagnostic in a semantic pass, don't fail the CS here
- Fail the CS, put the diagnostic in the error recovery part of the CS

PR #799, opened 2025-12-16
Feature: Clang Importer

### LINKED ISSUE #798: Implement ClangImporter

Snowy1803:
Closes #798 
ASan no longer works if the used LLVM version doesn't use asan, so ENABLE_ASAN can no longer be used if FROM_SOURCE is 0.
The macOS CI remains without them for faster builds.
The Linux CI LLVM19 remains without them as well.
The new Linux CI LLVM20 (Asserts + ASAN) builds LLVM with ASan, with a cache. It is relatively fast. The Linux version of ASan also has LeakSanitizer, so this one is more useful than the macOS one.

### DIFF
--- a/lib/ClangImporter/DeclImporter.cpp
+++ b/lib/ClangImporter/DeclImporter.cpp
@@ -1,8 +1,6 @@
 #include "DeclImporter.hpp"
 
 #include "AST/Attributes.hpp"
-#include "AST/Decls.hpp"
-#include "AST/Types.hpp"
 #include "Basic/SourceLocation.hpp"
 
 #include <clang/AST/Decl.h>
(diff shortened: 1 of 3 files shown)


lib/ClangImporter/DeclImporter.cpp
```diff
@@ -0,0 +1,192 @@
...
+#include <llvm/Support/Casting.h>
+
+namespace glu::clangimporter {
+
+bool DeclImporter::VisitFunctionDecl(clang::FunctionDecl *funcDecl)
+{
+    // Skip function definitions (we only want declarations)
+    if (funcDecl->hasBody()) {
+        return true;
+    }
```
LindonAliu: this means you can silently miss callable symbols I think

lib/ClangImporter/TypeConverter.cpp
```diff
@@ -0,0 +1,122 @@
...
+    clang::Type const *canonicalType
+        = clangType.getCanonicalType().getTypePtr();
+    if (auto type = _ctx.typeCache.lookup(canonicalType)) {
+        return type;
+    }
+
+    auto &typesArena = _ctx.glu.getTypesMemoryArena();
+    glu::types::TypeBase *gluType = nullptr;
+
+    if (auto *builtinType = llvm::dyn_cast<clang::BuiltinType>(canonicalType)) {
```
LindonAliu: This only resolves builtin/pointer/array/function types and relies on the cache for records/enums, so signatures become order‑dependent. If a function references a struct/enum declared later (or only forward‑declared), the function is skipped or its pointer becomes *Void and I think is an ABI‑breaking type mismatch

lib/ClangImporter/ClangImporter.cpp
```diff
@@ -1,15 +1,71 @@
...
+    // Run the tool with our custom action factory
+    ImportActionFactory factory(importCtx);
+    if (tool.run(&factory) != 0) {
+        llvm::errs() << "Failed to parse file: " << headerPath << "\n";
+        return nullptr;
+    }
+
+    // Create module declaration
+    auto *moduleDecl = glu::ast::ModuleDecl::create(
+        astContext.getASTMemoryArena().getAllocator(), SourceLocation::invalid,
```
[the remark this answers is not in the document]
LindonAliu: Implemented in #802

LindonAliu (approved):

PR #800, opened 2025-12-19
refactor: change drop function to take pointer types for structures

### LINKED ISSUE #792: Change drop function to take a pointer
For now when you overload the drop function of a struct, it takes a copy of the struct instead of a ptr which is heavy for nothing.

badermarouan:
Fix #792


Snowy1803:
This has to be remade change the GIL behavior instead of changing the IRGen behavior to avoid useless memory usage

Snowy1803 (dismissed):

PR #803, opened 2025-12-25
Feature: Explicit enum values

### LINKED ISSUE #786: Support explicit values for enum cases
Fix this TODO:
```
// CHECK:    EnumDecl {{0x[0-9a-f]+}} <line:18:1>
// CHECK-NEXT:      -->Name: MyEnum
// ## TODO: This should be 1 and 2, but = is not supported yet.
// CHECK:      -->Name: CASE1
// CHECK:      -->Name: CASE2
enum MyEnum: Int { CASE1 = 1, CASE2 = 2 }
```
The value of the cases should be reflected in IRGen::visitEnumVariantInst and in Debug Info.

LindonAliu:
Closes #786

### DIFF
--- a/lib/Sema/SemanticPass/EnumValueResolver.hpp
+++ b/lib/Sema/SemanticPass/EnumValueResolver.hpp
@@ -4,46 +4,48 @@
 #include "AST/ASTWalker.hpp"
 #include "Basic/Diagnostic.hpp"
 
-#include <llvm/ADT/APSInt.h>
-
 namespace glu::sema {
 
 /// @brief Resolves enum case values and ensures they have literal expressions.
 class EnumValueResolver : public ast::ASTWalker<EnumValueResolver, void> {
     DiagnosticManager &_diagManager;
 
-    static unsigned getEnumBitWidth(ast::EnumDecl *node)
+    /// @brief Gets the representable type for an enum, creating a default i32
+    /// if none specified.
+    /// @return The representable type (IntTy or CharTy).
+    static types::TypeBase *
+    getOrCreateRepresentableType(ast::EnumDecl *node, ast::ASTContext &ctx)
     {
         auto *repr = node->getRepresentableType();
-        if (!repr)
-            return 32;
+        if (repr)
+            return repr->getCanonicalType(ctx);
+        return ctx.getTypesMemoryArena().create<types::IntTy>(
+            types::IntTy::Signed, 32
+        );
+    }
 
-        auto *canonical
-            = repr->getCanonicalType(*node->getModule()->getContext());
-
-        if (auto *intTy = llvm::dyn_cast_or_null<types::IntTy>(canonical)) {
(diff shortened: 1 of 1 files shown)


lib/Sema/SemanticPass/EnumValueResolver.hpp
```diff
@@ -0,0 +1,120 @@
...
+namespace glu::sema {
+
+/// @brief Resolves enum case values and ensures they have literal expressions.
+class EnumValueResolver : public ast::ASTWalker<EnumValueResolver, void> {
+    DiagnosticManager &_diagManager;
+
+    struct EnumReprInfo {
+        unsigned bitWidth = 32;
+        bool isUnsigned = false;
+    };
```
Snowy1803: We don’t even support negative integer literals, this seems unnecessary. All that matters is the bitwidth. Past Sema (which is where those values are used), only the bit pattern matters, not the signedness or the sign.

lib/Sema/SemanticPass/EnumValueResolver.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+        return 32;
+    }
+
+    static ast::LiteralExpr *makeLiteral(
+        ast::ASTContext &ctx, llvm::APSInt const &value, SourceLocation loc
+    )
+    {
+        auto *litType
+            = ctx.getTypesMemoryArena().create<types::TypeVariableTy>();
```
Snowy1803: wtf, semantic passes shouldn’t have type variables

lib/Sema/SemanticPass/EnumValueResolver.hpp
```diff
@@ -0,0 +1,110 @@
...
+        bool hasValue = false;
+        llvm::APSInt current(bitWidth, false);
+        llvm::APSInt explicitValue(bitWidth, false);
+
+        for (auto *field : node->getFields()) {
+            auto *valueExpr = field->getValue();
+            if (valueExpr
+                && tryGetLiteralValue(valueExpr, bitWidth, explicitValue)) {
+                field->setValue(
+                    makeLiteral(*ctx, explicitValue, valueExpr->getLocation())
```
Snowy1803: The type of the literal is known here, it’s the representable type?

lib/Sema/SemanticPass/EnumValueResolver.hpp
```diff
@@ -0,0 +1,110 @@
...
+                current = explicitValue;
+                hasValue = true;
+                continue;
+            }
+
+            if (!hasValue) {
+                current = llvm::APSInt(bitWidth, false);
+                hasValue = true;
+            } else {
+                current += llvm::APSInt(llvm::APInt(bitWidth, 1), false);
```
Snowy1803: ++current; should work right

Snowy1803 (approved):

PR #805, opened 2026-01-12
Feature: Resolve templated struct content

### LINKED ISSUE #783: Resolve templated struct content
The use of a templated structure should be resolved — a parameterized StructTy should have its fields mapped to the concrete type:
```swift
struct Machin<T> {
    hello: T
}
func main() {
    let machin: Machin<Int> = { 3 };
    print(machin.hello);
}
```
There should be no `T` in the AST for the main, as it should be replaced by `Int`.

LindonAliu:
Closes #783 

### Description
This PR implements field type resolution for generic structs. When using `Wrapper<Int>`, fields that have type `T` are now correctly resolved to `Int`.

#### 1. Template parameter substitution via StructTy
Added `StructTy::getSubstitutedFieldType(index)` method that transforms `TemplateParamTy` into concrete types using the struct's template arguments. The `TemplateParamSubstituter` is now a local class within `StructTy.cpp`.

Used in:
* `ConstraintSystem.cpp`: for member access and struct initialization type checking
* `TypeLowering.hpp`: for LLVM IR and debug info generation

#### 2. Simplified type interning (AST)
`StructTy` instances with the same declaration and template arguments are interned (uniqued), so pointer comparison is now used in `HashType.cpp` for efficiency.

#### 3. Simplified buildLoad API (GILGen)
The `buildLoad`, `buildLoadCopy`, and `buildLoadTake` methods now automatically infer the type from the `PointerTy`, avoiding inconsistencies between the passed type and the actual pointer type.

### Tests
Added a functional test `test/functional/run/templated_struct_member.glu` that verifies:

* Accessing fields of generic structs
* Initializing generic structs
* Nested structs with generic types

### DIFF
--- a/include/AST/Types/StructTy.hpp
+++ b/include/AST/Types/StructTy.hpp
@@ -92,12 +92,6 @@ public:
         );
     }
 
-    /// @brief Returns the type of a field with template parameters substituted.
-    /// @param index The index of the field.
-    /// @return The field type with template parameters replaced by concrete
-    /// types.
-    TypeBase *getSubstitutedFieldType(size_t index);
-
     /// @brief Static method to check if a type is a StructTy.
     static bool classof(TypeBase const *type)
     {


lib/AST/HashType.cpp
```diff
@@ -65,9 +65,11 @@ class HashVisitor : public TypeVisitor<HashVisitor, std::size_t> {
...
     std::size_t visitStructTy(StructTy *type)
     {
-        auto templateHash = llvm::hash_combine_range(
-            type->getTemplateArgs().begin(), type->getTemplateArgs().end()
-        );
+        // Hash template args by their content, not by pointer values
+        std::size_t templateHash = 0;
+        for (auto *arg : type->getTemplateArgs()) {
+            templateHash = llvm::hash_combine(templateHash, arg->hash());
+        }
```
Snowy1803: pointer value is correct, that's the WHOLE POINT why they're uniqued. It's an optimization. USE IT.

PR #809, opened 2026-01-14
Feature: Auto import C, Rust, Zig

### LINKED ISSUE #779: Implement auto-importing C using clang
Add ability to import `.c` files without compiling them to bitcode
The glu import manager should be able to compile them to bitcode automatically, if it hasn't been done yet.
The compiler should launch `clang -g -c -emit-llvm file.c -o /tmp/something.bc`, then import the resulting bitcode file.
This should be done on last resort, if no precompiled bitcode file is found. The import manager should try all import paths for bitcode before trying all import paths for C code.

LindonAliu:
Closes #779

### DIFF
--- a/lib/IRDec/DITypeLifter.cpp
+++ b/lib/IRDec/DITypeLifter.cpp
@@ -162,10 +162,13 @@ public:
             return handleStructureType(diCompositeType);
         case llvm::dwarf::DW_TAG_array_type: {
             if (auto elements = diCompositeType->getElements()) {
-                if (elements->getNumOperands() != 2) {
+                if (elements->getNumOperands() == 0) {
                     return nullptr;
                 }
-                auto *subrange = llvm::dyn_cast<llvm::DISubrange>(elements[1]);
+                // Get the first (or only) subrange element
+                auto *subrange = llvm::dyn_cast<llvm::DISubrange>(
+                    elements[elements->getNumOperands() - 1]
+                );
                 if (!subrange) {
                     return nullptr;
                 }


lib/Sema/ImportManager.cpp
```diff
@@ -7,11 +7,26 @@
...
 namespace glu::sema {
 
+ImportManager::~ImportManager()
+{
+    for (auto const &entry : _generatedBitcodePaths) {
+        if (entry.second.empty()) {
+            continue;
+        }
+        llvm::sys::fs::remove(entry.second);
+    }
```
Snowy1803: never delete generated files, they're already in `/tmp` so they'll be deleted by the OS automatically
- macOS deletes after 3 days automatically
- Linux deletes after each reboot
Generated files are usually still needed after use, as the debug info is stored in them

lib/Sema/ImportManager.cpp
```diff
@@ -172,12 +196,185 @@ bool ImportManager::loadCHeader(SourceLocation importLoc, FileID fid)
...
+        llvm::SmallString<128> tempPath;
+        std::error_code ec
+            = llvm::sys::fs::createTemporaryFile("glu-import", "ll", tempPath);
+        if (ec) {
+            _diagManager.error(
+                importLoc,
+                "Failed to create temporary LLVM IR file: " + ec.message()
+            );
+            return false;
+        }
```
Snowy1803: a lot of duplicated code between C/Rust handling?

PR #810, opened 2026-01-14
Feature: Static Arrays For Loop

### LINKED ISSUE #742: Add support for iterating on an array
Support For Loops on static arrays (special case in Sema)

LindonAliu:
Closes #742 

### Description
This PR implements for-loop iteration over static arrays. When using `for value in arr` where `arr` is a `StaticArrayTy` (e.g., `Int[3]`), the loop now iterates over the array elements using inline pointer-based iteration.

#### 1. ForStmt array iteration flag (AST)
Added `isArrayIteration` flag to `ForStmt` to distinguish between iterator-based iteration (ranges) and inline pointer-based iteration (static arrays). The flag is printed in ASTPrinter for debugging.

#### 2. Array detection in Sema
Modified `LocalCSWalker::_visitForStmt()` to detect when the range expression is a `StaticArrayTy`:
- Added `getUnderlyingRangeType()` helper to look through `RefExpr` to get the declared variable type
- When array is detected, calls `setupArrayIteration()` which sets the flag and binds element type
- Otherwise uses existing iterator-based `setupIteratorIteration()` for ranges

Also added `BindToArrayElement` constraint kind for potential future use with type variables.

#### 3. GIL generation for array iteration
Added `GILGenStmt::visitForStmtArray()` which generates inline pointer-based iteration:
- Bitcasts array pointer to element pointer type
- Creates begin (first element) and end (past-last) pointers
- Loop condition: pointer equality comparison
- Dereferences current pointer for binding value
- Increments pointer with `buildPtrOffset`

#### 4. IRGen builtin function support
Modified `IRGen::visitCallInst()` to recognize synthetic builtin functions by name prefix `"builtin_"` in addition to the `BuiltinKind` enum, enabling the use of builtins created dynamically for array iteration.

#### 5. TypeVariableTyMapper update
Modified to skip visiting iterator functions (`begin`, `end`, `next`, `deref`, `==`) when `isArrayIteration` is true, since these are not used for array iteration.

### Tests
Added functional tests:
* `test/functional/Sema/for_stmt_array.glu` - Verifies AST shows `IsArrayIteration: true` for arrays
* `test/functional/run/for_array.glu` - Verifies runtime behavior:
  - Iterating over array elements
  - Accumulating values in a loop
  - Mixing array iteration with range iteration in the same function

### DIFF
--- a/lib/Sema/ConstraintSystem/ConstraintSystem.cpp
+++ b/lib/Sema/ConstraintSystem/ConstraintSystem.cpp
@@ -400,21 +400,15 @@ ConstraintResult ConstraintSystem::applyBindToArrayElement(
         }
 
         if (substitutedElement == staticArrayType->getDataType()) {
-            state.defaultableConstraintsSatisfied++;
-            return ConstraintResult::Applied;
+            return ConstraintResult::Satisfied;
         }
         // Try to unify first with the element type
         if (unify(elementType, staticArrayType->getDataType(), state)) {
-            state.defaultableConstraintsSatisfied++;
             return ConstraintResult::Applied;
         }
     }
 
-    // If the array type is still a type variable, we need to wait
-    if (llvm::isa<glu::types::TypeVariableTy>(substitutedArray)) {
-        return ConstraintResult::Failed;
-    }
-
+    // We don't have enough information to bind the array type
     return ConstraintResult::Failed;
 }
 
--- a/lib/Sema/ConstraintSystem/TypeVariableTyMapper.cpp
+++ b/lib/Sema/ConstraintSystem/TypeVariableTyMapper.cpp
@@ -41,14 +41,11 @@ public:
     {
         visit(node->getBinding());
         visit(node->getRange());
-        // Only visit iterator functions if not array iteration
-        if (!node->isArrayIteration()) {
-            visit(node->getBeginFunc());
-            visit(node->getEndFunc());
-            visit(node->getNextFunc());
-            visit(node->getDerefFunc());
-            visit(node->getEqualityFunc());
-        }
+        visit(node->getBeginFunc());
+        visit(node->getEndFunc());
+        visit(node->getNextFunc());
+        visit(node->getDerefFunc());
+        visit(node->getEqualityFunc());
         // Skip body
     }
 };


lib/IRGen/IRGen.cpp
```diff
@@ -604,10 +604,6 @@ struct IRGenVisitor : public glu::gil::InstVisitor<IRGenVisitor> {
     {
         auto callee = inst->getFunctionOrNull();
         assert(callee && "Built-in calls must have a named function");
-        auto builtin = callee->getDecl()->getBuiltinKind();
-        assert(
-            builtin != ast::BuiltinKind::None && "Function must be a built-in"
-        );
```
Snowy1803: shouldn't be removed

lib/Sema/ConstraintSystem/LocalCSWalker.cpp
```diff
@@ -241,9 +259,60 @@ class LocalCSWalker : public glu::ast::ASTWalker<LocalCSWalker, void> {
...
+                node, "builtin_eq", &ast::ForStmt::setEqualityFunc,
+                { uintType, uintType }, typesArena.create<glu::types::BoolTy>(),
+                { builtinsNs }
+            );
+            _cs.addConstraint(branches[0]);
+            return;
+        }
+
+        // For non-array types, the iterator constraints are added directly
+        // by setupIteratorIteration, so we don't need the disjunction
```
Snowy1803: ????

lib/Sema/ConstraintSystem/LocalCSWalker.cpp
```diff
@@ -241,9 +259,60 @@ class LocalCSWalker : public glu::ast::ASTWalker<LocalCSWalker, void> {
...
+        } else {
+            // For arrays, the BindToArrayElement constraint will handle
+            // everything. We also need to set up the EqualityFunc for pointer
+            // comparison in GILGen.
+            auto *uintType = typesArena.create<glu::types::IntTy>(
+                types::IntTy::Unsigned, 64
+            );
+            llvm::StringRef builtinsNs = "builtins";
+            createRangeAccessorRef(
+                node, "builtin_eq", &ast::ForStmt::setEqualityFunc,
```
Snowy1803: you can't setEqualityFunc with two different values in the different branches, as both branches must be evaluated

